PR #3, opened 2019-01-23
netdev-rte-offloads: bug fixes, add support for vxlan item

asafpenso:

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -150,6 +150,7 @@ struct rte_flow_items {
     struct rte_flow_item_udp udp;
     struct rte_flow_item_sctp sctp;
     struct rte_flow_item_icmp icmp;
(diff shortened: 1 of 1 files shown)


lib/netdev-rte-offloads.c
```diff
@@ -1010,7 +1011,7 @@ add_flow_rss_action(struct flow_actions *actions,
         rss->queue[i] = i;
     }
 
-    add_flow_action(actions, RTE_FLOW_ACTION_TYPE_RSS, rss);
+    //add_flow_action(actions, RTE_FLOW_ACTION_TYPE_RSS, rss);
```
OphirMunk: @asafpenso please explain why RSS action is removed

asafpenso: It causes a crash, which we are debugging. Currently, in order to proceed with vxlan decap, it was temporarily removed

PR #4, opened 2019-01-23
Read HW flow statistics

OphirMunk:
Add reading HW flow statistics

### DIFF
--- a/lib/dpif-netdev.c
+++ b/lib/dpif-netdev.c
@@ -522,6 +522,7 @@ struct dp_netdev_flow {
 
     bool dead;
     uint32_t mark;               /* Unique flow mark assigned to a flow */
+    bool is_hwol;                /* true if flow is fully offloaded */
 
     /* Statistics. */
     struct dp_netdev_flow_stats stats;
@@ -2371,6 +2372,7 @@ dp_netdev_flow_offload_put(struct dp_flow_offload_item *offload)
                           offload->actions_len, &flow->mega_ufid, &info,
                           NULL);
     ovs_mutex_unlock(&pmd->dp->hw_offload_mutex);
+    flow->is_hwol = info.is_hwol;
 
     if (ret) {
         if (!modification) {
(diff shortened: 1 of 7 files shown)


lib/dpif-netdev.c
```diff
@@ -3564,6 +3564,37 @@ dpif_netdev_flow_dump_thread_destroy(struct dpif_flow_dump_thread *thread_)
...
+    struct dp_netdev_port *port;
+    struct dpif_flow_stats stats;
+
+    odp_port_t in_port = netdev_flow->flow.in_port.odp_port;
+
+    ovs_mutex_lock(&pmd->dp->port_mutex);
+    port = dp_netdev_lookup_port(pmd->dp, in_port);
+    if (!port) {
+            ovs_mutex_unlock(&pmd->dp->port_mutex);
+            return -1;
```
asafpenso: Please add a print with the failure reason. Please do it for all other relevant places

lib/netdev-provider.h
```diff
@@ -867,6 +867,13 @@ struct netdev_class {
     int (*flow_del)(struct netdev *, const ovs_u128 *ufid,
                     struct dpif_flow_stats *);
 
+    /* Get offloaded flow stats. When a flow is fully offloaded, reading the
+     * PMD stats will result is no updates. In order to reflect the actual
```
asafpenso: typo?
"result is" -> "result in" / "result with"

lib/netdev.h
```diff
@@ -207,6 +207,7 @@ struct offload_info {
      * it will be in the pkt meta data.
      */
     uint32_t flow_mark;
+    bool is_hwol; /* true if flow is fully offloaded */
```
asafpenso: "fully offloaded" -> "fully offloaded to hw".
Also, why are you using the work "fully"? the flow can either be offloaded or not.

lib/netdev-dpdk.c
```diff
@@ -4069,6 +4068,32 @@ ufid_to_rte_flow_disassociate(const ovs_u128 *ufid) {
               UUID_ARGS((struct uuid *)ufid));
 }
 
+static int
+netdev_dpdk_flow_stats_get(struct netdev *netdev, const ovs_u128 *ufid,
```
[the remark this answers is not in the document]
rbaryanai: All the offload code it in rte-offloads.

lib/dpif-netdev.c
OphirMunk: Instead of print I add an assert condition 
```
assert(port);
```
due to the following reasons:

1. Error is not expected. If it exists it's a design error and not user configuration or alike
1. This function is called at a high rate so printing an error will overwhelmed the log file (which is unappropriate)

lib/netdev-provider.h
OphirMunk: Typo fixed in next PR: result "in"

lib/netdev.h
OphirMunk: Will update in next PR to: 
```
/* true if flow is offloaded to hardware */
```

lib/netdev-dpdk.c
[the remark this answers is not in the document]
OphirMunk: Fixed in next PR

PR #8, opened 2019-01-29
Hwol mlnx dpdk17.11

OphirMunk:
Implement flow aging and statistics

### DIFF
--- a/lib/dpif-netdev.c
+++ b/lib/dpif-netdev.c
@@ -3587,7 +3587,6 @@ dp_netdev_offload_used(struct dp_netdev_flow *netdev_flow,
             ovs_mutex_unlock(&pmd->dp->port_mutex);
             return -1;
     }
-    /* get offloaded stats */
     ret = netdev_flow_stats_get(port->netdev,
                     &netdev_flow->mega_ufid, &stats);
     ovs_mutex_unlock(&pmd->dp->port_mutex);
--- a/lib/dpif-netdev.c
+++ b/lib/dpif-netdev.c
@@ -3565,6 +3565,38 @@ dpif_netdev_flow_dump_thread_destroy(struct dpif_flow_dump_thread *thread_)
 }
 
 static int
+dp_netdev_offload_used(struct dp_netdev_flow *netdev_flow,
+                       struct dp_netdev_pmd_thread *pmd)
+{
+    int ret;
+    struct dp_netdev_port *port;
+    struct dpif_flow_stats stats;
+
+    odp_port_t in_port = netdev_flow->flow.in_port.odp_port;
+
+    ovs_mutex_lock(&pmd->dp->port_mutex);
+    port = dp_netdev_lookup_port(pmd->dp, in_port);
+    ovs_assert(port);
+    if (!port) {
+            ovs_mutex_unlock(&pmd->dp->port_mutex);
+            return -1;
+    }
+    /* get offloaded stats */
+    ret = netdev_flow_stats_get(port->netdev,
+                    &netdev_flow->mega_ufid, &stats);
+    ovs_mutex_unlock(&pmd->dp->port_mutex);
+    if (ret) {
+            return -1;
+    }
+    atomic_store_relaxed(&netdev_flow->stats.used,
+                    pmd->ctx.now / 1000);
+    non_atomic_ullong_add(&netdev_flow->stats.packet_count, stats.n_packets);
+    non_atomic_ullong_add(&netdev_flow->stats.byte_count, stats.n_bytes);
+
+    return 0;
+}
+
+static int
 dpif_netdev_flow_dump_next(struct dpif_flow_dump_thread *thread_,
                            struct dpif_flow *flows, int max_flows)
 {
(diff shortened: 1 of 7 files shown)


OphirMunk:
Pull request was updated. @asafpenso @rbaryanai please review

lib/netdev-dpdk.c
```diff
@@ -4069,6 +4068,35 @@ ufid_to_rte_flow_disassociate(const ovs_u128 *ufid) {
               UUID_ARGS((struct uuid *)ufid));
 }
 
+static int
+netdev_dpdk_flow_stats_get(struct netdev *netdev, const ovs_u128 *ufid,
```
rbaryanai: This function should be implemented in netdev-rte-offloads.c

lib/netdev-dpdk.c
```diff
@@ -4069,6 +4068,35 @@ ufid_to_rte_flow_disassociate(const ovs_u128 *ufid) {
...
+                     struct dpif_flow_stats *stats) {
+
+    int ret = 0;
+    struct rte_flow_error error;
+    struct rte_flow_query_count query;
+    struct netdev_dpdk *dev = netdev_dpdk_cast(netdev);
+
+    struct rte_flow *rte_flow = ufid_to_dpdk_rte_flow(ufid);
```
rbaryanai: might have many RTE flows, correct only in the default case.

lib/netdev-rte-offloads.c
```diff
@@ -692,6 +692,29 @@ int netdev_vport_flow_del(struct netdev * netdev OVS_UNUSED,
...
+    rte_port = netdev_rte_port_search(port_no, &dpdk_map);
+
+    if (rte_port == NULL) {
+        VLOG_ERR("failed to find dpdk port for port %d",port_no);
+        return NULL;
+    }
+
+    struct ufid_hw_offload *ufid_hwol = ufid_hw_offload_find(ufid, &rte_port->ufid_to_rte);
+    if (ufid_hwol) {
+        return ufid_hwol->rte_flow_data[0].flow;
```
rbaryanai: correct only if we have one rte_flow.

lib/netdev-rte-offloads.h
```diff
@@ -53,4 +53,6 @@ int netdev_dpdk_flow_put(struct netdev *netdev, struct match *match,
 int netdev_dpdk_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
                      struct dpif_flow_stats *stats OVS_UNUSED);
 
+struct rte_flow *
+ufid_to_dpdk_rte_flow(const ovs_u128 *ufid);
```
rbaryanai: should not be exposed.

lib/netdev-rte-offloads.c
```diff
@@ -692,6 +692,29 @@ int netdev_vport_flow_del(struct netdev * netdev OVS_UNUSED,
     return 0;
 }
 
+struct rte_flow *
+ufid_to_dpdk_rte_flow(const ovs_u128 *ufid) {
```
asafpenso: probably should be static since it's not defined in header file

lib/netdev-dpdk.c
```diff
@@ -4069,6 +4068,35 @@ ufid_to_rte_flow_disassociate(const ovs_u128 *ufid) {
...
+    if (!rte_flow) {
+        return EINVAL;
+    }
+    memset(&query, 0, sizeof(query));
+    /* reset counters after query */
+    query.reset = 1;
+    ret = rte_flow_query(dev->port_id, rte_flow, RTE_FLOW_ACTION_TYPE_COUNT,
+                         &query, &error);
+    if (ret) {
```
asafpenso: I suggest doing if (!ret) stats->n_packets etc... 
This way you'll have a single return ret.

lib/dpif-netdev.c
```diff
@@ -3564,6 +3564,38 @@ dpif_netdev_flow_dump_thread_destroy(struct dpif_flow_dump_thread *thread_)
...
+    odp_port_t in_port = netdev_flow->flow.in_port.odp_port;
+
+    ovs_mutex_lock(&pmd->dp->port_mutex);
+    port = dp_netdev_lookup_port(pmd->dp, in_port);
+    ovs_assert(port);
+    if (!port) {
+            ovs_mutex_unlock(&pmd->dp->port_mutex);
+            return -1;
+    }
+    /* get offloaded stats */
```
asafpenso: the code is self explanatory so i think the comment can be omitted

lib/netdev-rte-offloads.c
OphirMunk: It is defined in netdev-rte-offloads.h

lib/netdev-dpdk.c
OphirMunk: Thanks but it seems more readable this way and I prefer not to indent a big block of code. This style is also exercised throughout the file so the current code is the correct style.

lib/dpif-netdev.c
OphirMunk: Agreed. Fixed.

PR #9, opened 2019-02-03
netdev-rte-offloads: use bitmap for checking actions existance

asafpenso:
When parsing nl_actions it is essential to know which actions we have.
This can be used to decide on the correct flow path. For example:
which action we need to add, whether it is a fate action, or whether
there are no actions and drop should be added to the flow

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -1422,7 +1422,9 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
     const struct nlattr *a;
     unsigned int left;
 
-    bool is_tunnel_pop_action = false;
+    /* actions in nl_actions will be asserted in this bitmap,
+     * according to their values in ovs_action_attr enum */
+    uint64_t is_action_bitmap = false;
     struct rte_flow_action_jump jump;
     struct rte_flow_action_count count;
 
@@ -1434,7 +1436,7 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
             if (!vport) {
                 break;
             }
-            is_tunnel_pop_action = true;
+            is_action_bitmap |= 1 << OVS_ACTION_ATTR_TUNNEL_POP;
             *counter_id = count.id;
             result = 0;
         } else {
@@ -1461,7 +1463,7 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
 
         /* If action is tunnel pop, create another table with a default flow.
          * Do it only once, if default rte flow doesn't exist */
-        if (is_tunnel_pop_action &&
+        if ((is_action_bitmap & (1 << OVS_ACTION_ATTR_TUNNEL_POP)) &&
                           !rte_port->default_rte_flow[vport->table_id]) {
             /* The default flow has the lowest priority, no pattern
              * (match all) and Mark action */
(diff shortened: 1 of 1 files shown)


rbaryanai:
looks fine

lib/netdev-rte-offloads.c
```diff
@@ -1422,7 +1422,9 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
     const struct nlattr *a;
     unsigned int left;
 
-    bool is_tunnel_pop_action = false;
+    /* actions in nl_actions will be asserted in this bitmap,
+     * according to their values in ovs_action_attr enum */
+    uint64_t is_action_bitmap = false;
```
OphirMunk: For a bitmap variable should assign 0 rather that "false"

lib/netdev-rte-offloads.c
```diff
@@ -1212,10 +1210,20 @@ prepare_and_add_jump_count_flow_action(
     jump->group = rte_port->table_id;
     add_flow_action(actions, RTE_FLOW_ACTION_TYPE_JUMP, jump);
 
+    return rte_port;
+}
+
+static void
+prepare_and_add_count_flow_action(
+        const struct nlattr *nlattr,
```
OphirMunk: You are calling with an unused parameter: nlattr. Please remove it.

lib/netdev-rte-offloads.c
```diff
@@ -1212,10 +1210,20 @@ prepare_and_add_jump_count_flow_action(
     jump->group = rte_port->table_id;
     add_flow_action(actions, RTE_FLOW_ACTION_TYPE_JUMP, jump);
 
+    return rte_port;
+}
+
+static void
+prepare_and_add_count_flow_action(
```
OphirMunk: Dito: I suggest renaming this function: prepare_and_add_count_flow_action ==> netdev_rte_add_flow_action

lib/netdev-rte-offloads.c
```diff
@@ -1191,13 +1191,11 @@ netdev_dpdk_validate_flow(const struct match *match) {
 }
 
 static struct netdev_rte_port *
-prepare_and_add_jump_count_flow_action(
+prepare_and_add_jump_flow_action(
```
OphirMunk: All functions names in this file should have some relevance to netdev-rte-offload. Specifically for this function I suggest renaming this function: prepare_and_add_jump_flow_action ==> netdev_rte_jump_flow_action

lib/netdev-rte-offloads.c
asafpenso: done

lib/netdev-rte-offloads.c
```diff
@@ -1434,7 +1436,7 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
             if (!vport) {
                 break;
             }
-            is_tunnel_pop_action = true;
+            is_action_bitmap |= 1 << OVS_ACTION_ATTR_TUNNEL_POP;
```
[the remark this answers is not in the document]
asafpenso: it's too small indeed, but an infra change which is required. for baidu the commit arrangements is not critical

lib/netdev-rte-offloads.c
asafpenso: done

lib/netdev-rte-offloads.c
asafpenso: Since it's a static function i dont think it matters, but i'll change

lib/netdev-rte-offloads.c
asafpenso: Since it's a static function i dont think it matters, but i'll change

PR #13, opened 2019-02-10
netdev-rte-offloads: fix vxlan rte flows priority

evelineraine:
Signed-off-by: Eveline Raine <evelineraine@gmail.com>

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -1478,7 +1478,7 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
              * (match all) and Mark action */
             const struct rte_flow_attr def_flow_attr = {
                 .group = vport->table_id,
-                .priority = 0, /* lowest priority */
+                .priority = 1,
                 .ingress = 1,
                 .egress = 0,
             };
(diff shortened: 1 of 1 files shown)


asafpenso:
merged

lib/netdev-rte-offloads.c
```diff
@@ -1847,7 +1847,7 @@ netdev_vport_vxlan_add_rte_flow_offload(struct netdev_rte_port * rte_port,
 
     const struct rte_flow_attr flow_attr = {
         .group = rte_port->table_id,
-        .priority = 1,
+        .priority = 0,
```
OphirMunk: We should not set hard coded priorities in code. We should dynamically translate OVS flow priorities into DPDK priorities. We should also reserve an internal rte priority range for internal usage. For example the default TAG action should have a low priority and this priority should be hard coded (but probably lower than 1).

evelineraine: Do we have priorities for OVS dataplane flows? I haven't seen it in dpif/dump-flows.
Also, from DPDK docs, we get that not all hardware can support multiple priorities. I'm not sure how much hardware DPDK runs on supports our offloads, but not multiple priorities.

asafpenso: We'll look into that once we work on the upstream code.

PR #14, opened 2019-02-11
Fix offloaded VXLAN aging and fix UDP port masks

evelineraine:


lib/netdev-dpdk.c
```diff
@@ -1094,6 +1094,16 @@ int  netdev_dpdk_get_port_id(const struct netdev * netdev)
     return CONTAINER_OF(netdev, struct netdev_dpdk, up)->port_id;
 }
 
+int  netdev_dpdk_reset_hw_counters(struct netdev *netdev)
+{
```
asafpenso: I think we should add a null check for the parameter and return -1 is null.
If the convention is not doing so, then the function can return void.

evelineraine: Done

asafpenso:
@OphirMunk please review Eveline's code. I'd like to merge it.

lib/netdev-rte-offloads.c
```diff
@@ -1664,7 +1789,10 @@ netdev_dpdk_flow_del(struct netdev *netdev OVS_UNUSED, const ovs_u128 *ufid,
         netdev_rte_port_ufid_hw_offload_free(ufid_hw_offload );
     }
 
-    return -1;
+    dev->hw_n_packets = 0;
```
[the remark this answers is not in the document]
asafpenso: This is already happening. Please see the latest commit inside this PR.

PR #15, opened 2019-02-13
ofproto-dpif: Add offloaded:yes to dpif/dump-flow

OphirMunk:
When executing command: ovs-appctl dpif/dump-flows <bridge name> and
in case a flow is offloaded - add an "offloaded:yes" indication.


asafpenso:
Merged

PR #20, opened 2019-02-19
netdev-rte-offloads: Extract pattern matching handling to a function.

asafpenso:
Handling of flows pattern matching is being done as part of
netdev_dpdk_add_rte_flow_offload function.

Extracting this logic to a dedicate function provides better
readability and an option to reuse that logic in a different
place.

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -367,18 +367,18 @@ struct action_rss_data {
 
 static struct action_rss_data *
 add_flow_rss_action(struct flow_actions *actions,
-                    struct netdev *netdev) {
+                    uint16_t num_queues) {
     int i;
     struct action_rss_data *rss_data;
 
     rss_data = xmalloc(sizeof(struct action_rss_data) +
-                       sizeof(uint16_t) * netdev->n_rxq);
+                       sizeof(uint16_t) * num_queues);
     *rss_data = (struct action_rss_data) {
         .conf = (struct rte_flow_action_rss) {
             .func = RTE_ETH_HASH_FUNCTION_DEFAULT,
             .level = 0,
             .types = 0,
-            .queue_num = netdev->n_rxq,
+            .queue_num = num_queues,
             .queue = rss_data->queue,
             .key_len = 0,
             .key  = NULL
@@ -386,7 +386,7 @@ add_flow_rss_action(struct flow_actions *actions,
     };
 
     /* Override queue array with default */
-    for (i = 0; i < netdev->n_rxq; i++) {
+    for (i = 0; i < num_queues; i++) {
        rss_data->queue[i] = i;
     }
 
@@ -575,7 +575,7 @@ netdev_dpdk_add_rte_flow_offload(struct netdev *netdev,
     add_flow_action(&actions, RTE_FLOW_ACTION_TYPE_MARK, &mark);
 
 
-    rss = add_flow_rss_action(&actions, netdev);
+    rss = add_flow_rss_action(&actions, netdev->n_rxq);
     add_flow_action(&actions, RTE_FLOW_ACTION_TYPE_END, NULL);
 
     flow = netdev_dpdk_rte_flow_create(netdev, &flow_attr,patterns.items,
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -603,7 +603,7 @@ add_flow_patterns(struct flow_patterns *patterns,
     return 0;
 }
 
-static int
+static struct rte_flow *
 netdev_dpdk_add_rte_flow_offload(struct netdev *netdev,
                                  const struct match *match,
                                  struct nlattr *nl_actions OVS_UNUSED,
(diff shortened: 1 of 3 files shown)


asafpenso:
Additional commit was added:
dpif-netdev: Update value of MAX_FLOW_MARK.

asafpenso:
Added 2 more commits for review.
Please disregard the WIP commit.

lib/netdev-rte-offloads.c
```diff
@@ -589,40 +620,24 @@ netdev_dpdk_add_rte_flow_offload(struct netdev *netdev,
...
     struct flow_items spec, mask;
 
     memset(&spec, 0, sizeof spec);
     memset(&mask, 0, sizeof mask);
 
-    ret = add_flow_patterns(&patterns, &spec, &mask, match);
-    if (ret) {
+    result = add_flow_patterns(&patterns, &spec, &mask, match);
+    if (result) {
         flow = NULL;
```
rbaryanai: flow is already null

lib/netdev-rte-offloads.c
```diff
@@ -573,6 +573,37 @@ netdev_rte_port_search_by_port_no(odp_port_t port_no)
     return rte_port;
 }
 
+static struct rte_flow *
+netdev_rte_offload_mark_rss(struct netdev *netdev,
+                            struct offload_info *info,
+                            struct flow_patterns *patterns,
+                            struct flow_actions *actions,
+                            struct rte_flow_action_mark *mark,
+                            struct action_rss_data *rss,
```
rbaryanai: No need to pass rss here, you allocation it inside here anyway.

lib/netdev-rte-offloads.c
```diff
@@ -573,6 +573,37 @@ netdev_rte_port_search_by_port_no(odp_port_t port_no)
...
+    add_flow_action(actions, RTE_FLOW_ACTION_TYPE_END, NULL);
+
+    flow = netdev_dpdk_rte_flow_create(netdev, flow_attr, patterns->items,
+                                       actions->actions, &error);
+
+    free(rss);
+    if (!flow) {
+        VLOG_ERR("%s: rte flow create offload error: %u : message : %s\n",
+                netdev_get_name(netdev), error.type, error.message);
+        flow = NULL;
```
rbaryanai: no need. should return NULL on failure anyway. error should use rate limit.

lib/netdev-rte-offloads.c
```diff
@@ -573,6 +573,37 @@ netdev_rte_port_search_by_port_no(odp_port_t port_no)
     return rte_port;
 }
 
+static struct rte_flow *
+netdev_rte_offload_mark_rss(struct netdev *netdev,
+                            struct offload_info *info,
+                            struct flow_patterns *patterns,
+                            struct flow_actions *actions,
+                            struct rte_flow_action_mark *mark,
```
rbaryanai: you have flow mark inside info. since you don't need to use it outside. no sense in passing it.
should be local

lib/netdev-rte-offloads.c
```diff
@@ -1221,8 +1133,10 @@ netdev_dpdk_flow_put(struct netdev *netdev, struct match *match,
         return -1;
     }
 
-    ufid_hw_offload_add_rte_flow(
-                ufid_hw_offload, rte_flow, netdev);
+    ufid_hw_offload_add_rte_flow(ufid_hw_offload, rte_flow, netdev);
```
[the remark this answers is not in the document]
asafpenso: I'm not fully following you...
We do netdev_dpdk_add_rte_flow_offload and if rte_flow is null we return -1. if it's not null than we call ufid_hw_offload_add_rte_flow.
So ufid is saved only if rte_flow is not null.

lib/netdev-rte-offloads.c
```diff
@@ -603,6 +604,98 @@ add_flow_patterns(struct flow_patterns *patterns,
...
+    rte_port = netdev_rte_port_search(port_no,  &vport_map);
+    if (rte_port == NULL) {
+        rte_port = netdev_rte_port_search(port_no,  &dpdk_map);
+    }
+
+    return rte_port;
+}
+
+static struct netdev_rte_port *
+netdev_rte_add_jump_flow_action(
```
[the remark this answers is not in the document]
asafpenso: I see your point. 
Currently, the function hides all the details of getting the table id out of rte_port.
With your suggestion, there is no point to this function since if it takes num as a parameter is would look like:
    jump->group = num;
    add_flow_action(actions, RTE_FLOW_ACTION_TYPE_JUMP, jump);
So, i prefer to hide the details and once we implement CT, we can think about the solution.

lib/netdev-rte-offloads.c
asafpenso: mark is the actual parameter that i need to pass to rte, so it's needed. 
Maybe you meant that flow_mark can be put into mark outside this function and in that case no need to pass info to the function. Is that correct?

lib/netdev-rte-offloads.c
```diff
@@ -634,10 +692,39 @@ netdev_dpdk_add_rte_flow_offload(struct netdev *netdev,
...
-    struct action_rss_data *rss = NULL;
-    flow = netdev_rte_offload_mark_rss(netdev, info, &patterns, &actions,
-                                       &mark, rss, &flow_attr);
+    const struct nlattr *a = NULL;
+    unsigned int left = 0;
+    struct rte_flow_action_jump jump = {0};
+    struct rte_flow_action_count count = {0};
+    struct netdev_rte_port *vport = NULL;
+
+    NL_ATTR_FOR_EACH_UNSAFE (a, left, nl_actions, actions_len) {
```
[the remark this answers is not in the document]
asafpenso: Once we have the capabilities, if needed, i'll rebase accordingly. 
Can we still do something currently?

lib/netdev-rte-offloads.c
asafpenso: Done

asafpenso:
@rbaryanai @OphirMunk I handled Roni's comment and uploaded new patch set. Please review these commits.

asafpenso:
@rbaryanai @OphirMunk I handled Roni's comment and uploaded new patch set. Please review these commits.

asafpenso:
@rbaryanai 
Please review the last 10 commits from 71966ea to 3f0ac44.
I handled all of your comments, and uploaded additional commits

PR #21, opened 2019-02-24
Add marked statistics

OphirMunk:


asafpenso:
Looks ok. Waiting for @rbaryanai  to review as well before i merge this

lib/dpif-netdev.c
```diff
@@ -6244,6 +6266,9 @@ packet_batch_per_flow_execute(struct packet_batch_per_flow *batch,
     dp_netdev_flow_used(flow, batch->array.count, batch->byte_count,
                         batch->tcp_flags, pmd->ctx.now / 1000);
 
+    dp_netdev_flow_hwol(flow, batch->flow->mark,
```
rbaryanai: Why you don't update the statistics where flow is found by mark? in the following code:
 dp_packet_has_flow_mark(packet, &mark)) {
            flow = mark_to_flow_find(pmd, mark);
            if (OVS_LIKELY(flow)) {
                tcp_flags = parse_tcp_flags(packet);

Why passing this information and do it later. emc also updated inside the function although it is not per flow.

lib/dpif-netdev.c
```diff
@@ -6254,7 +6279,7 @@ static inline void
 dp_netdev_queue_batches(struct dp_packet *pkt,
                         struct dp_netdev_flow *flow, uint16_t tcp_flags,
                         struct packet_batch_per_flow *batches,
-                        size_t *n_batches)
+                        size_t *n_batches, bool marked)
```
rbaryanai: same as previous, I don't think there is a need to do it here. to many code changes for passing something that you can update while you used it,

lib/dpif-netdev.c
OphirMunk: As we discussed - the place where the flow is found cannot be used to do mark++ for several reasons among which:
1. You update the statistics per batch where you accumulate counters
2. Keeping software engineering the same - all packets stats are handled in one function and are spread in the code.

rbaryanai: I don't agree, because you add mark to the code flow where mark might be irrelevant, but not going to insist as long as I can see that flow as offloaded.

OphirMunk: Please be specific where "you add mark to the code flow where mark might be irrelevant".
If you refer to the line:    
```
dp_netdev_flow_hwol(flow, batch->flow->mark,
```
Then please note the implementation:
```
static void
dp_netdev_flow_hwol(struct dp_netdev_flow *netdev_flow, uint32_t mark,
                    int mark_count)
{
    if (mark != INVALID_FLOW_MARK) {
        non_atomic_ullong_add(&netdev_flow->stats.mark_count, mark_count);
    }
}
```
Please note:
* In case of full offload: we will not pass here and mark count will not be incremented.
* In case hw-offload is disabled: mark should be invalid therefore mark count will not be incremented.
* In case of partial offload: we do pass here and mark should be valid, therefore mark count should be incremented.

I plan checking it carefully to be sure.

OphirMunk: Latest PR works for me (tested 2nd and 3rd cases).

asafpenso:
PR will be closed and not merged. The PR for baidu_upstream will be merged instead

PR #22, opened 2019-02-26
dpif: Add marked packets stats

OphirMunk:
When running datapath dump-flow command  with "-m" (more verbosity)
"ovs-appctl dpif/dump-flows -m <bridge-name>" add printing of the
number of marked packets.


asafpenso:
Merged.

PR #24, opened 2019-03-10
HW offload - infra patches by Roni

asafpenso:


asafpenso:
@OphirMunk  I rebased Roni's patches on top of upstream as of today.

PR #26, opened 2019-03-17
IPv6 Inner/Outer Hardware Offload

evelineraine:


yuskul:
bot:retest

yuskul:
bot:retest

yuskul:
bot:retest

lib/netdev-rte-offloads.c
```diff
@@ -1289,6 +1287,29 @@ add_dpdk_flow_patterns(struct flow_patterns *patterns,
                 masks->ipv4.hdr.next_proto_id;
     }
 
+    /* IP v6 */
+    if (match->flow.dl_type == htons(ETH_TYPE_IPV6)) {
+        memset(&specs->ipv6, 0, sizeof(specs->ipv6));
```
asafpenso: For now, it's good. Once you prepare the patches for the community need to remove the () from sizeof.
E.g: instead of sizeof(specs->ipv6) write sizeof specs->ipv6.

lib/netdev-rte-offloads.c
```diff
@@ -1289,6 +1287,29 @@ add_dpdk_flow_patterns(struct flow_patterns *patterns,
...
+        memset(&masks->ipv6, 0, sizeof(masks->ipv6));
+
+        specs->ipv6.hdr.proto = match->flow.nw_proto;
+        specs->ipv6.hdr.hop_limits = match->flow.nw_ttl;
+        rte_memcpy(specs->ipv6.hdr.src_addr, match->flow.ipv6_src.s6_addr, sizeof(specs->ipv6.hdr.src_addr));
+        rte_memcpy(specs->ipv6.hdr.dst_addr, match->flow.ipv6_dst.s6_addr, sizeof(specs->ipv6.hdr.dst_addr));
+
+        masks->ipv6.hdr.proto = match->wc.masks.nw_proto;
+        masks->ipv6.hdr.hop_limits = match->wc.masks.nw_ttl;
+        rte_memcpy(masks->ipv6.hdr.src_addr, match->wc.masks.ipv6_src.s6_addr, sizeof(masks->ipv6.hdr.src_addr));
```
asafpenso: Have you run checkpatch? looks like a long line that exceeds 80 chars.

lib/netdev-rte-offloads.c
```diff
@@ -1762,7 +1783,29 @@ add_vport_vxlan_flow_patterns(struct flow_patterns *patterns,
...
+        rte_memcpy(masks->ipv6.hdr.dst_addr, match->wc.masks.tunnel.ipv6_dst.s6_addr, sizeof(masks->ipv6.hdr.dst_addr));
+
+        add_flow_pattern(patterns, RTE_FLOW_ITEM_TYPE_IPV6,
+                         &specs->ipv6, &masks->ipv6);
+
+        /* Save proto for L4 protocol setup */
+        proto = specs->ipv6.hdr.proto &
+                masks->ipv6.hdr.proto;
     } else {
+        VLOG_ERR("tunnel L3 header is neither IPv4 nor IPv6");
```
asafpenso: If there was not a print before, we should keep it without it. In case of an error the terminal would be spammed with messages and reduced the perf. 
So, either use VLOG_ERR_RL which Roni added to limit the rate, or remove the print.

lib/netdev-rte-offloads.c
```diff
@@ -1401,8 +1400,11 @@ netdev_dpdk_add_rte_flow_offload(struct netdev_rte_port *rte_port,
     int result = -1;
     struct netdev_rte_port * vport = NULL;
 
+    VLOG_DBG("Adding rte offload for flow ufid "UUID_FMT, UUID_ARGS((struct uuid *) ufid));
```
asafpenso: I think all prints start with lower case letter, please confirm

lib/netdev-rte-offloads.c
```diff
@@ -1738,7 +1738,7 @@ add_vport_vxlan_flow_patterns(struct flow_patterns *patterns,
 
     /* IP v4 */
     uint8_t proto = 0;
-    if (match->flow.dl_type == htons(ETH_TYPE_IP)) {
+    if (match->wc.masks.tunnel.ip_src || match->wc.masks.tunnel.ip_dst) {
```
asafpenso: Can you explain which issue you resolved here and why existing code is not good?

evelineraine: This code runs on already decapped packet. When looking at flow.dl_type you are looking at the inner L3 protocol. This code block needs to determine the outer L3 protocol.
Since tunnel struct does not have tunnel L3 type, we look at which masks are set (for IPv4 src/dst or for IPv6 src/dst (in the following patches)).

lib/netdev-rte-offloads.c
evelineraine: In the netdev_dpdk_validate_flow() we are already checking which offload patterns we do and don't support. After this check, all invalid patterns for *outer* header are filtered out. This leaves us with either IPv4 or IPv6 as outer L3 protocols. If for some reason here we get neither, this is a real error.
`VLOG_ERR_RL` would be an option. Any reason we shouldn't make this a default and use everywhere in place of `VLOG_ERR`?

evelineraine:
> Please see my comments.
> In general, make sure to run checkpatch and update the code accordingly.

Yes, I actually forgot to do that.

> Besides the commit header, need also to add text to the body, giving more info on the patch itself.
> For now, it's ok and once you prepare the patches for the community need to take that into consideration.
> The commit header should start with a capital letter after the ':' and should end with '.'.

Ok, though in general the first line of commit message shouldn't end with period, and there isn't any note about this in OVS contributing guidelines. But I see _most_ commits end with a period, so it's a convention I didn't know about, I'll be following it.

I don't suppose there is a way to amend a commit inside a PR and not lose reviewers' comments?

asafpenso:
With our newly code we indeed use VLOG_ERR_RL.

Regards,
Asaf Penso

asafpenso:
Ophir and I discussed that in the past and we prefer to update the patches themselves, and not adding a new patch to fix the comments.
Afterwards to push -f , which will maintain the comments for the previous commits and will show the new patches as well.
I hope it is clear 😊

Regards,
Asaf Penso

OphirMunk:
Comments regarding commit: ("dpif-netdev: Improve ufid logging")

1. Same comment as before (and for the remaining commits): commit message body should be filled in.
A title only is not sufficient.

2. A comment added inside the code 

3. Please fix checkpatch warnings:


== Checking cc409080a341 ("dpif-netdev: Improve ufid logging") ==
WARNING: Line is 81 characters long (recommended limit is 79)
#20 FILE: lib/dpif-netdev.c:2155:
VLOG_WARN("Masked megaflow ufid "UUID_FMT" is not associated with a mark?\n",

WARNING: Line is 117 characters long (recommended limit is 79)
#38 FILE: lib/dpif-netdev.c:2187:
VLOG_DBG("Associated dp_netdev flow ufid "UUID_FMT" with mark %u", UUID_ARGS((struct uuid *) &flow->ufid), mark);

Lines checked: 45, Warnings: 2, Errors: 0

lib/dpif-netdev.c
```diff
@@ -2184,7 +2184,7 @@ mark_to_flow_associate(const uint32_t mark, struct dp_netdev_flow *flow)
                 hash_int(mark, 0));
     flow->mark = mark;
 
-    VLOG_DBG("Associated dp_netdev flow %p with mark %u\n", flow, mark);
+    VLOG_DBG("Associated dp_netdev flow ufid "UUID_FMT" with mark %u", UUID_ARGS((struct uuid *) &flow->ufid), mark);
```
OphirMunk: Why did you drop end of line "\n"? If by mistake - need to restore.

lib/netdev-rte-offloads.c
```diff
@@ -898,7 +898,7 @@ dump_flow_pattern(struct rte_flow_item *item)
 
         VLOG_DBG("rte flow ipv4 pattern:\n");
         if (ipv4_spec) {
-            VLOG_DBG("  Spec: tos=0x%"PRIx8", ttl=%"PRIx8", proto=0x%"PRIx8
+            VLOG_DBG("  Spec: tos=0x%"PRIx8", ttl=0x%"PRIx8", proto=0x%"PRIx8
```
OphirMunk: In OVS ttl is printed as unsigned char. So the fix should be:
```
 VLOG_DBG("  Spec: tos=0x%"PRIx8", ttl=%"PRIu8", proto=0x%"PRIx8
```

lib/netdev-rte-offloads.c
```diff
@@ -909,7 +909,7 @@ dump_flow_pattern(struct rte_flow_item *item)
             VLOG_DBG("  Spec = null\n");
         }
         if (ipv4_mask) {
-            VLOG_DBG("  Mask: tos=0x%"PRIx8", ttl=%"PRIx8", proto=0x%"PRIx8
+            VLOG_DBG("  Mask: tos=0x%"PRIx8", ttl=0x%"PRIx8", proto=0x%"PRIx8
```
OphirMunk: Same comment here regarding ttl

lib/netdev-rte-offloads.c
OphirMunk: When adding VLOG messages on an expected positive case we may be overwhelmed with printouts. For example during high insertion rate. Instead of VLOG printouts we should take care of adding statistics that will inform of the successful operation.

lib/netdev-rte-offloads.c
```diff
@@ -1812,6 +1814,9 @@ netdev_vport_vxlan_add_rte_flow_offload(struct netdev_rte_port * rte_port,
                                         struct offload_info * info,
                               struct dpif_flow_stats * flow_stats  OVS_UNUSED)
 {
+    VLOG_DBG("Adding rte offload for vport vxlan flow ufid "UUID_FMT,
+        UUID_ARGS((struct uuid *) ufid));
+
```
OphirMunk: Similar comment here regarding adding VLOG printouts on expected positive cases

lib/dpif-netdev.c
evelineraine: Actually VLOG_* already adds `\n` for messages automatically if not present. I will restore it, but it does not matter if we add `\n` or not.

lib/netdev-rte-offloads.c
evelineraine: Without this printout, I discovered, it is difficult to determine the location of errors during offloading. Besides, we are printing multiple debug messages with all the flow patterns we add. How does adding this message make it worse?

lib/netdev-rte-offloads.c
evelineraine: *(Same as above)*

lib/netdev-rte-offloads.c
evelineraine: Some start with an uppercase letter. For example see `lib/dpif-netdev.c#2351`. Is there a convention we follow?

evelineraine: I also checked `lib/netdev-dpdk.c`. There, all log messages start with an uppercase letter. Since our `lib/netdev-rte-offloads.c` branched from it, we probably should follow the uppercase convention for consistency.

lib/netdev-rte-offloads.c
evelineraine: *Fixed in the new patches*

lib/netdev-rte-offloads.c
evelineraine: *Fixed in the new patches*

lib/netdev-rte-offloads.c
evelineraine: Changed to VLOG_ERR_RL in new commit

lib/dpif-netdev.c
evelineraine: *Fixed in the new patches*

lib/netdev-rte-offloads.c
evelineraine: *Fixed in the new patches*

lib/netdev-rte-offloads.c
evelineraine: *Fixed in the new patches*

evelineraine:
> Regarding commit: ("packets: Add IPv6 printf pattern")
> 
> 1. Commit title: printf --> printout
> 2. Should elaborate in the commit message body about the contents of
>    the commit. A title only is not sufficient.
> 3. This is a preceding commit to another one which is several commits
>    away: "netdev-rte-offloads: Add flow dumping for IPv6".
>    The two commits should be consecutive.
> 4. Please note in file lib/packets.c that there is already an API for
>    printing IPv6 addresses:
> 
> ```
> /* Stores the string representation of the IPv6 address 'addr' into the
>  * character array 'addr_str', which must be at least INET6_ADDRSTRLEN
>  * bytes long. If addr is IPv4-mapped, store an IPv4 dotted-decimal string. */
>  const char *
>  ipv6_string_mapped(char *addr_str, const struct in6_addr *addr)
> ```
> Let's keep the same format in our code. Can you use this API for Baidu
> code?

1-3: Fixed 
#### 4: 
I know this function exists. However there is another approach - to use `*_FMT` macros. I think this has a better precedent - we use this approach to print UUIDs in log messages like so:
```c
VLOG_WARN("Masked megaflow ufid "UUID_FMT
                  " is not associated with a mark?\n",
                  UUID_ARGS((struct uuid *)mega_ufid));
```

In addition to this we already have such formatting macros in `lib/packets.h` for IPv4:
```c
#define IP_FMT "%"PRIu32".%"PRIu32".%"PRIu32".%"PRIu32
#define IP_ARGS(ip)                             \
    ntohl(ip) >> 24,                            \
    (ntohl(ip) >> 16) & 0xff,                   \
    (ntohl(ip) >> 8) & 0xff,                    \
    ntohl(ip) & 0xff
```

So this makes sense to me to add parity for IPv6.

However the two approaches differ in that `*_FMT` macros always prints 39 characters (zeroes or not) but `ipv6_string_mapped()` can compress addresses.

> Comments regarding commit: ("dpif-netdev: Improve ufid logging")
> 
> 1. Same comment as before (and for the remaining commits): commit message body should be filled in.
>    A title only is not sufficient.
> 2. A comment added inside the code
> 3. Please fix checkpatch warnings:
> 
> == Checking cc409080a341 ("dpif-netdev: Improve ufid logging") ==
> WARNING: Line is 81 characters long (recommended limit is 79)
> #20 FILE: lib/dpif-netdev.c:2155:
> VLOG_WARN("Masked megaflow ufid "UUID_FMT" is not associated with a mark?\n",
> 
> WARNING: Line is 117 characters long (recommended limit is 79)
> #38 FILE: lib/dpif-netdev.c:2187:
> VLOG_DBG("Associated dp_netdev flow ufid "UUID_FMT" with mark %u", UUID_ARGS((struct uuid *) &flow->ufid), mark);
> 
> Lines checked: 45, Warnings: 2, Errors: 0

1,3: Fixed

evelineraine:
bot:retest

OphirMunk:
> > ```
> > /* Stores the string representation of the IPv6 address 'addr' into the
> >  * character array 'addr_str', which must be at least INET6_ADDRSTRLEN
> >  * bytes long. If addr is IPv4-mapped, store an IPv4 dotted-decimal string. */
> >  const char *
> >  ipv6_string_mapped(char *addr_str, const struct in6_addr *addr)
> > ```
> > Let's keep the same format in our code. Can you use this API for Baidu
> > code?
> 
> 
> 
> #### 4:
> 
> I know this function exists. However there is another approach - to use `*_FMT` macros. I think this has a better precedent - we use this approach to print UUIDs in log messages like so:
> 
> 
> ```c
> VLOG_WARN("Masked megaflow ufid "UUID_FMT
>                   " is not associated with a mark?\n",
>                   UUID_ARGS((struct uuid *)mega_ufid));
> ```

There is a difference between ufid and IPv6 addresses. The former is condense. Every bit counts. We want to see all bits. The later is very sparse. We want to see only the few non-zero bits. This is why a standard was defined for printing IPv6 addresses in RFC 4291. https://tools.ietf.org/html/rfc4291. 
The API used in OVS for printing IPv6 addresses is 
```
ipv6_string_mapped()
```
and it is used dozens of times and uniquely in the OVS code. Can you please use it?
We should follow the OVS community standards. This is one of them.

> In addition to this we already have such formatting macros in `lib/packets.h` for IPv4:
> 
> ```c
> #define IP_FMT "%"PRIu32".%"PRIu32".%"PRIu32".%"PRIu32
> #define IP_ARGS(ip)                             \
>     ntohl(ip) >> 24,                            \
>     (ntohl(ip) >> 16) & 0xff,                   \
>     (ntohl(ip) >> 8) & 0xff,                    \
>     ntohl(ip) & 0xff
> ```

The end result is that IPv4 is printed according to the IPv4 standard even though it uses this formatting macros. We are not sparse for IPv4. Still for IPv6 we need to be sparse.

> However the two approaches differ in that `*_FMT` macros always prints 39 characters (zeroes or not) but `ipv6_string_mapped()` can compress addresses.

Yes...in order to be sparse.

OphirMunk:
> Changes the print format from `PRIx8` to `PRIu8`.
>

Whenever there is a fix for a previous commit, just before Signed-off-by ... at the end of the commit message you should add:
```
Fixes: <12 characters hash> ("<commit title>"). 
```
For example:
```
Fixes: b10d46a60013 ("tests: Check dpif-netdev odp_actions consistency.")
Signed-off-by: Ian Stokes <ian.stokes@intel.com>
```
You can find the fixed commit details by running 'git blame "file_name"', then receive the fixed commit SHA1 number, then run 'git show "SHA1 number"' for the full details

OphirMunk:
> Previously the code looked at decapsulated packet's Ethernet type to
> determine tunnel L3 protocol is IPv4. The fix involves checking tunnel
> address masks for IPv4 source and destination. If non-zero, the protocol
> is IPv4.

Same comment as before: there is a fix here, so should add "Fixes: ..." at the end of commit message just before "Signed-of-by..." line.

OphirMunk:
> OphirMunk 2 days ago  Member
> When adding VLOG messages on an expected positive case we may be overwhelmed with printouts. 
> For example during high insertion rate. Instead of VLOG printouts we should take care of adding
> statistics that will inform of the successful operation.
> 
> evelineraine 8 hours ago  Author Member
> Without this printout, I discovered, it is difficult to determine the location of errors during offloading. 

Have you checked an option to enhance error messages during offloading such that you it will be easy to determine the location of errors during offloading?

> Besides, we are printing multiple debug messages with all the flow patterns we add. How does adding 
> this message make it worse?

One of our goals is to be best in flow insertion rate. VLOG_DBG adds additional code (even with a low log level and no printouts) and may reduce the insertion rate (should be measured and verified). If you are right and there are multiple VLOG_DBG on the fast path - they should all be removed as well (but don't do it yourself now). Any VLOG_DBG that is on  an error path or on a non-fast path - is OK. We are only measured on successful fast path. 
By the way, looking at netdev-dpdk.c before it was split into netdev-rte-offloads.c - I notice no VLOG_DBG on the fast path.

OphirMunk:
> packets: Add IPv6 printout pattern.
>
As mentioned previously we should use
```
ipv6_string_mapped()
```
for printing IPv6 addresses. Therefore I suggest omitting this commit.

evelineraine:
@OphirMunk 
I have changed flow dumping to print compresses IPv6 addresses using `ipv6_string_mapped()` and the removed the commit that adds `*_FMT` patterns.
I have also added the `Fixes` tags to my commits. Thank you for pointing that out!

I still don't agree about debug prints. In our code (in addition to multiline pattern dumps) we have this message printed for each flow being offloaded:
```c
VLOG_DBG("second flow of decap created");
```
Without it, we wouldn't be able to debug all the issues we did on virtio since February. We have this line, same as the lines I added, to increase visibility into what is actually happening inside OVS. 
It does impact performance, of course. In the same way counting marked and hardware offloaded packets that you added does as well. But it was done for the same reason - to increase visibility.

Regarding the fast path vs slow path - flows offloading is not really fastpath, just like inserting rules into the kernel module in OVS-kernel is not fastpath either. I don't think it makes sense to optimize it as hard as to drop debug information in order to save cycles. In `dpif-netdev.c` several debug messages are being printed for each flow even before it reaches `netdev-rte-offloads.c`. 
If we want to squeeze the last bit of performance (while sacrificing some visibility in live production environments), we should add a way to compile-out debug messages with a `#define` flag like DPDK does.

All this being said, we really want to have this PR merged ASAP. If you still think I should remove the prints, I will and then we can follow up on this discussion later.

OphirMunk:
Merged into hwol-mlnx-dpdk17.11 branch

PR #32, opened 2019-04-10
Ovs upstream for review

OphirMunk:


lib/netdev-rte-offloads.c
```diff
@@ -731,3 +732,28 @@ netdev_rte_offloads_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
...
+ * In case of vport class and "vxlan" type we update it to match netdev
+ * datapath.
+ */
+void
+netdev_rte_offloads_add_port(struct netdev * netdev)
+{
+    const char *type = netdev_get_type(netdev);
+    if (netdev_vport_is_vport_class(netdev->netdev_class)) {
+            if (!strcmp("vxlan", type)) {
+            netdev_rte_offload_vxlan_init(netdev);
```
asafpenso: Ident

lib/netdev-rte-offloads.c
```diff
@@ -731,3 +732,28 @@ netdev_rte_offloads_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
...
+ * This function is called as part of adding a new dpif netdev port.
+ * In case of vport class and "vxlan" type we update it to match netdev
+ * datapath.
+ */
+void
+netdev_rte_offloads_add_port(struct netdev * netdev)
+{
+    const char *type = netdev_get_type(netdev);
+    if (netdev_vport_is_vport_class(netdev->netdev_class)) {
+            if (!strcmp("vxlan", type)) {
```
asafpenso: ident

lib/netdev-rte-offloads.c
```diff
@@ -731,3 +732,28 @@ netdev_rte_offloads_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
...
+    cls->init_flow_api = NULL;
+}
+
+/*
+ * This function is called as part of adding a new dpif netdev port.
+ * In case of vport class and "vxlan" type we update it to match netdev
+ * datapath.
+ */
+void
+netdev_rte_offloads_add_port(struct netdev * netdev)
```
asafpenso: What about adding else cases for below "if"s?

lib/netdev-rte-offloads.c
```diff
@@ -731,3 +732,28 @@ netdev_rte_offloads_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
...
+
+    cls->flow_put = NULL;
+    cls->init_flow_api = NULL;
+}
+
+/*
+ * This function is called as part of adding a new dpif netdev port.
+ * In case of vport class and "vxlan" type we update it to match netdev
+ * datapath.
+ */
```
asafpenso: Shouldn't this be in a header file?
What's the coding convention for documentation? Doxygen?

lib/netdev-rte-offloads.h
```diff
@@ -54,4 +54,6 @@ int netdev_rte_offloads_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
     .flow_put = netdev_rte_offloads_flow_put,   \
     .flow_del = netdev_rte_offloads_flow_del
 
+void netdev_rte_offloads_add_port(struct netdev * netdev);
```
asafpenso: remove space between "*" and "netdev"

asafpenso: looking now at other functions looks like the verb is at the end, so maybe consider having port_add instead of add_port.

lib/dpif-netdev.c
```diff
@@ -52,6 +52,7 @@
 #include "netdev.h"
 #include "netdev-provider.h"
 #include "netdev-vport.h"
+#include "netdev-rte-offloads.h"
```
asafpenso: should be above netdev-vport.h according to alphabet

lib/netdev-rte-offloads.c
```diff
@@ -731,3 +732,28 @@ netdev_rte_offloads_flow_del(struct netdev *netdev, const ovs_u128 *ufid,
 
     return netdev_rte_offloads_destroy_flow(netdev, ufid, rte_flow);
 }
+
+static void
+netdev_rte_offload_vxlan_init(struct netdev *netdev)
+{
+    struct netdev_class *cls = (struct netdev_class *)netdev->netdev_class;
+
```
rbaryanai: I think we should make sure all function pointers are set, and also we should save previous addresses and restore them when port is removed.

asafpenso:
Not relevant anymore

PR #46, opened 2019-07-01
Vxlan fixes v2

elibritstein:
v1-v2: thanks Ophir
- better commit order - shorter and split a commit
- possible memory corruption bug fix


lib/netdev-rte-offloads.c
```diff
@@ -922,28 +895,45 @@ netdev_rte_offloads_add_flow(struct netdev *netdev,
...
             goto out;
         }
 
+        if (action_bitmap & (1 << OVS_ACTION_ATTR_CLONE)) {
+            struct flow_actions jump_actions = { .actions = NULL, .cnt = 0 };
+
+            jump.group = 1;
+            add_flow_action(&jump_actions, RTE_FLOW_ACTION_TYPE_JUMP, &jump);
+            add_flow_action(&jump_actions, RTE_FLOW_ACTION_TYPE_END, NULL);
+
```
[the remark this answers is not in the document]
elibritstein: right. will be fixed

lib/netdev-rte-offloads.c
```diff
@@ -1072,8 +1075,7 @@ netdev_rte_offloads_flow_put(struct netdev *netdev, struct match *match,
 
     if (ufid_hw_offload) {
         VLOG_DBG("got modification and destroying previous rte_flow");
-        ufid_hw_offload_remove(ufid, &rte_port->ufid_to_rte);
-        ret = netdev_rte_port_ufid_hw_offload_free(ufid_hw_offload);
+        ret = netdev_offloads_flow_del(ufid);
```
OphirMunk: Function netdev_offloads_flow_del() always return positive errors. Therefore the test 
``
if (ret < 0)
`` is non effective. More than that - the function netdev_offloads_flow_del() ignores any errors from netdev_rte_port_ufid_hw_offload_free(). It must be fixed to consider those errors.

elibritstein:
> Commit "netdev-rte-offloads: Fix VXLAN encap flows" is a WA which should be reverted as soon as SW steering fixes table #0 limitations. Do you confirm that his commit only contains the WA and no other functionalities?

It cannot be reverted as the first implementation was also a WA, but a wrong one. It should be amended to avoid the jump rule when sw-str fixes it. it contains only this functionality

lib/netdev-rte-offloads.c
elibritstein: right. though netdev_rte_port_ufid_hw_offload_free returns always 0...
it should be done in another commit though, maybe in other placed in the code as well.
i'd suggest to comply with the standard and return negative value as an error

lib/netdev-rte-offloads.c
```diff
@@ -1394,8 +1394,10 @@ netdev_vport_vxlan_add_rte_flow_offload(struct netdev_rte_port *rte_port,
             }
 
             if (flow) {
+                ufid_hw_offload_add(ufid_hw_offload, &rte_port->ufid_to_rte);
+                ufid_to_portid_add(ufid, rte_port->dp_port);
```
OphirMunk: Adding OVS flow (per ufid) should be done unconditionally to dpdk flow. Calling the 2 lines above should occur outside of the loop. Attached is a suggested patch
```
diff --git a/lib/netdev-rte-offloads.c b/lib/netdev-rte-offloads.c
index b7a2425..9bd9691 100644
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -1492,6 +1492,9 @@ netdev_vport_vxlan_add_rte_flow_offload(struct netdev_rte_port *rte_port,
         return -1;
     }
 
+    ufid_hw_offload_add(ufid_hw_offload, &rte_port->ufid_to_rte);
+    ufid_to_portid_add(ufid, rte_port->dp_port);
+
     struct rte_flow_attr flow_attr = {
         .group = rte_port->table_id,
         .priority = 0,
```

PR #51, opened 2019-07-04
Ipv6 vxlan

elibritstein:
note that for outer ipv6 vxlan, need pmd fix in
https://github.com/elibritstein/dpdk.org/tree/ipv6-zero-udp-csum

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -811,7 +811,7 @@ netdev_rte_offloads_add_flow(struct netdev *netdev,
                              const struct match *match,
                              struct nlattr *nl_actions,
                              size_t actions_len,
-                             const ovs_u128 *ufid OVS_UNUSED,
+                             const ovs_u128 *ufid,
                              struct offload_info *info,
                              struct rte_flow **rte_flow0)
 {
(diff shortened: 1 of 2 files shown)


elibritstein:
the pmd patch on top of 19.05 is at:
http://patches.dpdk.org/patch/56072/

the pmd patch on top of master-net-mlx is at:
http://patches.dpdk.org/patch/56081/

lib/netdev-rte-offloads.c
```diff
@@ -1177,9 +1177,9 @@ add_vport_vxlan_flow_patterns(struct flow_patterns *patterns,
 
     /* IP v4 */
     uint8_t proto = 0;
-    if (match->flow.dl_type == htons(ETH_TYPE_IP)) {
-        memset(&spec->ipv4, 0, sizeof spec->ipv4);
-        memset(&mask->ipv4, 0, sizeof mask->ipv4);
+    if (match->wc.masks.tunnel.ip_src || match->wc.masks.tunnel.ip_dst) {
+        memset(&spec->ipv4, 0, sizeof(spec->ipv4));
+        memset(&mask->ipv4, 0, sizeof(mask->ipv4));
```
OphirMunk: By coding conventions sizeof should not use braces. The original code (without braces) should be restored.

elibritstein: what coding conventions do you refer?
according to OVS' one, it passes utilities/checkpatch.py
also, please see:
https://news.ycombinator.com/item?id=9629461
https://stackoverflow.com/questions/5894892/why-and-when-do-i-need-to-use-parentheses-after-sizeof
and also the examples in:
https://wikileaks.org/ciav7p1/cms/page_26607644.html

elibritstein:
> The commit message does not explain what was the original bug and how this patch fixes it.
> Need to update it.

if you refer to the one in dpdk-pmd, the commit message was clear enough to dpdk reviewers, and it is already applied to master-net-mlx branch.

lib/netdev-rte-offloads.c
OphirMunk: Let's merge this PR. It's not that urgent. But FYI the coding convention of OVS is here:
https://github.com/openvswitch/ovs/blob/master/Documentation/internals/contributing/coding-style.rst
And the sizeof relevant paragraph is:
```
When the operand of sizeof is an expression, there is no need to parenthesize that operand, and please don't.
```

elibritstein: done

PR #52, opened 2019-07-07
Hw stats

elibritstein:


lib/netdev-rte-offloads.c
```diff
@@ -1120,6 +1120,70 @@ netdev_rte_offloads_flow_put(struct netdev *netdev, struct match *match,
...
+            if (dpdk_port_id != DPDK_ETH_PORT_ID_INVALID) {
+                memset(&query, 0, sizeof query);
+                /* reset counters after query */
+                query.reset = 1;
+                ret = rte_flow_query(dpdk_port_id, rte_flow,
+                                     actions.actions, &query, &error);
+                if (ret) {
+                    continue;
+                }
+                stats->n_packets += (query.hits_set) ? query.hits : 0;
```
rbaryanai: what is the purpose of if here? why not just adding the result.

lib/netdev-rte-offloads.c
```diff
@@ -1120,6 +1120,70 @@ netdev_rte_offloads_flow_put(struct netdev *netdev, struct match *match,
...
+        if (rte_flow) {
+            dpdk_port_id = netdev_dpdk_get_port(netd);
+            if (dpdk_port_id != DPDK_ETH_PORT_ID_INVALID) {
+                memset(&query, 0, sizeof query);
+                /* reset counters after query */
+                query.reset = 1;
+                ret = rte_flow_query(dpdk_port_id, rte_flow,
+                                     actions.actions, &query, &error);
+                if (ret) {
+                    continue;
```
rbaryanai: maybe we need a DBG log here, in case it fails.

lib/netdev-rte-offloads.c
```diff
@@ -1120,6 +1120,70 @@ netdev_rte_offloads_flow_put(struct netdev *netdev, struct match *match,
...
+    /*
+     * Several HW flows may be related for one ufid. For example for one vport
+     * flow we may have created a HW flow per each physical port. Therefore we
+     * need to loop over all HW flows related to this ufid and accumulate the
+     * statitistics related to it.
+     */
+    for (int i = 0 ; i < uho->curr_idx ; i++) {
+        netd = uho->rte_flow_data[i].netdev;
+        rte_flow = uho->rte_flow_data[i].flow;
+        if (rte_flow) {
```
rbaryanai: get_stats and flow_put/flow_del are called from different context. 
what if between calls we have flow destroy? hope it won't crash the system.

lib/dpif-netdev.c
```diff
@@ -545,6 +545,7 @@ struct dp_netdev_flow {
 
     bool dead;
     uint32_t mark;               /* Unique flow mark assigned to a flow */
+    bool is_hwol;                /* true if flow is fully offloaded */
```
rbaryanai: maybe as mentioned in comment we should specify not-offloaded, mark-rss, fully offload.
if we want indication about if flow was offloaded or not, seems like this is the place and we would like to know if flow was offloaded even just with mark-rss.

lib/netdev-rte-offloads.c
elibritstein: if there is an error i don't think we should trust the result is valid, or still zeroed.
it is needed as there are flows, like encap, that have 2 rules, one with count and another one without (the workaround).

lib/netdev-rte-offloads.c
elibritstein: it may occur normally, as i wrote above, but ok

lib/netdev-rte-offloads.c
elibritstein: we can add a lock, but i'm not sure it can happen. i think it won't be needed after we rebase and get 5fc5c50f3 ("netdev: Dynamic per-port Flow API."). what do you think?

lib/netdev-rte-offloads.c
rbaryanai: I meant the query.hit_set? , understand the ret.

lib/netdev-rte-offloads.c
rbaryanai: right.

lib/netdev-rte-offloads.c
rbaryanai: need to dig into the code. basically  the dpif the defend by ref count, so if you query a flow you add ref count to the flow. same for for the put/del flow, if flow is deleted while reading stats it will be removed only when ref count get to zero. for offload we are not protected because the execution order is async and we don't hold ref count.

lib/dpif-netdev.c
OphirMunk: The flag 'is_hwol' which is documented as "is fully offloaded" was determined in the early stage of development. It was very confusing at that time since for most people MARK+RSS was actually an offloaded operation as well. But now it becomes completely obsolete as we even split actions in the middle (e.g. in connection tracking misses) between HW and SW.
I suggest renaming 'is_hwol' as 'hwcnt' (HW counter) which means whether we read counters from HW or from SW. This is a simple and exact description of this flag and it will not change as development continues.

lib/dpif-netdev.c
```diff
@@ -3596,6 +3596,36 @@ dpif_netdev_flow_dump_thread_destroy(struct dpif_flow_dump_thread *thread_)
...
+    /* get offloaded stats */
+    ret = netdev_flow_stats_get(port->netdev, &netdev_flow->mega_ufid, &stats);
+    ovs_mutex_unlock(&pmd->dp->port_mutex);
+    if (ret) {
+        return -1;
+    }
+    atomic_store_relaxed(&netdev_flow->stats.used, pmd->ctx.now / 1000);
+    non_atomic_ullong_add(&netdev_flow->stats.packet_count, stats.n_packets);
+    non_atomic_ullong_add(&netdev_flow->stats.byte_count, stats.n_bytes);
+
```
OphirMunk: I wonder if we should comment here and add to documentation the fact that we only read from HW the number of packets and bytes, but not the TCP flags which remain 0 in the stats struct. This is a known limitation when reading from HW. Maybe better letting the reviewers raise this issue first.

lib/netdev-rte-offloads.c
OphirMunk: Some more contentions scenarios:
* Physical ports are added/deleted while vport statistics are dumped. In order to test we can run a background script that adds/deletes ports while dumping the vport statistics continuously.
* Having multiple vports. In that case they all share the same physical ports. In order to test we can define multiple vports while simultaneously dumping their stats.

OphirMunk: Eli - why do you think that the refactored code 5fc5c50f3 ("netdev: Dynamic per-port Flow API.") can change the contentions risks? It's about dynamically selecting the correct vport on startup and will probably behave the same during run time.

elibritstein: i mistakenly thought the functions are of the flow, but they are of the netdev.
i added a lock for the stats/remove.
regarding netdev adds/removes, we are already protected

lib/dpif-netdev.c
elibritstein: note that i fixed it to do this only if there are n_packets
it had a bug that used was updated even for no packets.
regarding the add - it's just useless if there are no n_packets.

lib/netdev-rte-offloads.c
elibritstein: i think hits should be treated as valid only if hits_set is set

PR #53, opened 2019-07-09
netdev-rte-offloads: Support non-encap forwarding flows

elibritstein:
SW steering does not support output (forwarding) on table #0. Apply the
workaround done for encap flows (clone action) also for non-encap
forwarding flows (output action) as well.

Issue: 1820646
Signed-off-by: Eli Britstein <elibr@mellanox.com>

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -931,16 +931,18 @@ netdev_rte_offloads_add_flow(struct netdev *netdev,
         VLOG_DBG("Flow with Mark and RSS actions: NIC offload was %s",
                  flow ? "succeeded" : "failed");
     } else {
-        /* Table 0 does not support encap. Set the encap action in table #1,
-         * and the same matches and jump to table #1 in table #0.
-         * This is bad for performance and insertion rate but as a WA for
-         * SW-STR.
+        /* Table 0 does not support forwarding (SW steering limitation). As a
+         * WA set the output action in table #1 with the same matches from
+         * table #0 and replace the output action from table #0 with a jump
+         * action to table #1.
+         * This is bad...
          */
 
         /* Actions are supported, offload the flow */
         flow_attr.transfer = 1;
-        /* The flows for encap should be added to group 1 */
-        if (action_bitmap & (1 << OVS_ACTION_ATTR_CLONE)) {
+        /* The flows for output should be added to group 1 */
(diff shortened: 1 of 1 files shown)


lib/netdev-rte-offloads.c
```diff
@@ -931,16 +931,17 @@ netdev_rte_offloads_add_flow(struct netdev *netdev,
         VLOG_DBG("Flow with Mark and RSS actions: NIC offload was %s",
                  flow ? "succeeded" : "failed");
     } else {
-        /* Table 0 does not support encap. Set the encap action in table #1,
-         * and the same matches and jump to table #1 in table #0.
+        /* Table 0 does not support forwarding. Set the forwarding action in
+         * table #1, and the same matches and jump to table #1 in table #0.
```
OphirMunk: I would rephrase the comment as follows:
```
  /* Table 0 does not support forwarding. As a WA set the output action in
   * table #1 with the same matches from table #0 and replace the output 
   * action from table #0 with a jump action to table #1.
   * This is bad...
   */

File: lib/netdev-rte-offloads.c [hunk 2]
```diff
@@ -931,16 +931,17 @@ netdev_rte_offloads_add_flow(struct netdev *netdev,
...
          * SW-STR.
          */
 
         /* Actions are supported, offload the flow */
         flow_attr.transfer = 1;
-        /* The flows for encap should be added to group 1 */
-        if (action_bitmap & (1 << OVS_ACTION_ATTR_CLONE)) {
+        /* The flows for output should be added to group 1 */
+        if (action_bitmap & (1 << OVS_ACTION_ATTR_CLONE) ||
+            action_bitmap & (1 << OVS_ACTION_ATTR_OUTPUT)) {
```
OphirMunk: Why do we need to set the CLONE action in the bitmap? It currently hides the real internal actions: TUNNEL_PUSH and OUTPUT which should have been set in action bitmap (but unfortunately are not). Then in the lines above you should only check for the OUTPUT action. 
This is also correct to do since the CLONE action may hide other actions than OUTPUT, and then the WA should not be applicable.

lib/netdev-rte-offloads.c
elibritstein: ok

lib/netdev-rte-offloads.c
elibritstein: as this is a workaround that will be kicked out, i won't change it for now

PR #54, opened 2019-07-15
forwarder TSO

noaezra:
Support TCP segmentation offload (TSO) in virtio-forwarder. The TCP is a
technique for increasing egress throughput by passing a multipacket
buffer to the NIC, while the NIC splits the buffer into separate
packets. This technique reduces CPU overhead.

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -2099,6 +2099,12 @@ netdev_rte_offloads_port_init(uint16_t relay_id,
     rte_eth_dev_info_get(port, &dev_info);
     netdev_rte_offloads_port_conf.rxmode.offloads = 0;
 
+    /* enable tso for vf: */
+    netdev_rte_offloads_port_conf.txmode.offloads = 0;
+    if (!vm)
+        netdev_rte_offloads_port_conf.txmode.offloads = (DEV_TX_OFFLOAD_TCP_TSO |
+                DEV_TX_OFFLOAD_MULTI_SEGS);
+
     ret = rte_eth_dev_configure(port, queue_num, queue_num,
             &netdev_rte_offloads_port_conf);
     if (ret < 0) {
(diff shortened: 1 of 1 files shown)


lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
+
+    if (l4_proto_id == IPPROTO_TCP) {
+        ol_flags |= (PKT_TX_TCP_SEG | PKT_TX_TCP_CKSUM);
+        m->l2_len = l2_len;
+        m->l3_len = l3_len;
+        m->l4_len = l4_len;
+        m->ol_flags = ol_flags;
+        ret = rte_eth_dev_get_mtu(port_id, &mtu);
+        if (ret < 0) {
+            mtu = 1500;
```
elibritstein: 1500 -> ETHERMTU

lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
+    struct tcp_hdr *tcp_hdr;
+    uint32_t l2_len = 0;
+    uint32_t l3_len = 0;
+    uint32_t l4_len = 0;
+    uint64_t ol_flags = 0;
+    uint8_t l4_proto_id = 0;
+    uint16_t mtu = 0;
+    int ret = 0;
+
+    if (m->pkt_len < 1400)
```
elibritstein: what is 1400? a finger number? if so, maybe
#define FORWARDER_MIN_TSO_MTU  (ETHERMTU - 100)
and use this const

lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
 }
 
+static void
+parse_packet_fields(struct rte_mbuf *m, uint16_t port_id)
+{
+    struct ipv4_hdr *ipv4_hdr;
+    struct ipv6_hdr *ipv6_hdr;
+    struct ether_hdr *eth_hdr;
+    struct tcp_hdr *tcp_hdr;
+    uint32_t l2_len = 0;
```
elibritstein: avoid unnecessary initializations, for example if not IPv4/IPv6 packet

lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
+    if (m->pkt_len < 1400)
+        return;
+
+    eth_hdr = rte_pktmbuf_mtod(m, struct ether_hdr *);
+    l2_len = sizeof(struct ether_hdr);
+
+    switch (rte_be_to_cpu_16(eth_hdr->ether_type)) {
+    case ETHER_TYPE_IPv4:
+        ipv4_hdr = (struct ipv4_hdr *) ((char *)eth_hdr + l2_len);
+        l3_len = (ipv4_hdr->version_ihl & 0x0f) * 4;
```
elibritstein: 4 -> sizeof(uint32_t)

lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
+        l4_proto_id = ipv4_hdr->next_proto_id;
+        if (l4_proto_id == IPPROTO_TCP) {
+            tcp_hdr = (struct tcp_hdr *)((char *)ipv4_hdr + l3_len);
+            l4_len = (tcp_hdr->data_off & 0xf0) >> 2;
```
elibritstein: >> 2 --> / sizeof(uint32_t)

lib/netdev-rte-offloads.c
rbaryanai: use bit shift (...& 0xf) << 2

lib/netdev-rte-offloads.c
rbaryanai: you should always extract the eth and check that (pkt->len - eth_len)  < MTU

lib/netdev-rte-offloads.c
rbaryanai: tcp_header->data_off >> 4 gives you the size of the header in 4 bytes units. so 5 means 20 bytes. 
so it should be << 2 . together it is >>2 which is ok (after the & 0xf0)

lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
+        ipv6_hdr = (struct ipv6_hdr *) ((char *)eth_hdr + l2_len);
+        l4_proto_id = ipv6_hdr->proto;
+        if (l4_proto_id == IPPROTO_TCP) {
+            l3_len = sizeof(struct ipv6_hdr);
+            tcp_hdr = (struct tcp_hdr *)((char *)ipv6_hdr + l3_len);
+            l4_len = (tcp_hdr->data_off & 0xf0) >> 2;
+            ol_flags |= PKT_TX_IPV6;
+        }
+        break;
+    default:
```
rbaryanai: if we have vlan it wont work. which is fine, but it is a problem if don't say anything.

lib/netdev-rte-offloads.c
```diff
@@ -2209,14 +2215,77 @@ netdev_rte_offloads_update_relay(uint16_t relay_id,
...
+            tcp_hdr = (struct tcp_hdr *)((char *)ipv4_hdr + l3_len);
+            l4_len = (tcp_hdr->data_off & 0xf0) >> 2;
+            ol_flags |= (PKT_TX_IPV4 | PKT_TX_IP_CKSUM);
+        }
+        break;
+    case ETHER_TYPE_IPv6:
+        ipv6_hdr = (struct ipv6_hdr *) ((char *)eth_hdr + l2_len);
+        l4_proto_id = ipv6_hdr->proto;
+        if (l4_proto_id == IPPROTO_TCP) {
+            l3_len = sizeof(struct ipv6_hdr);
```
rbaryanai: this is true only if we don't have extensions. ipv6 extension is a pain, but we need to support in the future.

lib/netdev-rte-offloads.c
elibritstein: i think /4 is compiled to >>2, so we can do the code more readable

lib/netdev-rte-offloads.c
elibritstein: i think /4 is compiled to >>2, so we can do the code more readable

lib/netdev-rte-offloads.c
elibritstein: i agree, but not in the code.
if/when we support VLAN/IPv6-ext, add to the code to support.
currently, it is handled ok (not doing TSO)

lib/netdev-rte-offloads.c
elibritstein: and *4 compiled to <<2

rbaryanai: compiler will optimize for sure. I usually don't like to see * and / in a per packet code. not really important.

lib/netdev-rte-offloads.c
rbaryanai: If user machine is configured with it, all packets will be dropped and there is not indication why. so it is not handled ok.

PR #55, opened 2019-07-16
Ct i1 fixes

elibritstein:


elibritstein:
we can have the same functionality of the hw-offload=debug without touching the code, but use DPDK shared objects, and use LD_PRELOAD, so maybe we should not use it

PR #63, opened 2019-07-31
Ct nat

elibritstein:


OphirMunk:
1. All in all - the NAT bug fixes and the support for IPv6 - look good.  👍 +1

2. Do we have CT support for IPv6? Not sure since I see that the other fillers do not support IPv6. For example:
* conntrack_off_fill_key();
* conntrack_off_fill_match();
* conntrack_off_fill_tun();

Is this correct?

elibritstein:
> 1. All in all - the NAT bug fixes and the support for IPv6 - look good.  👍 +1
> 2. Do we have CT support for IPv6? Not sure since I see that the other fillers do not support IPv6. For example:
> 
> * conntrack_off_fill_key();
> * conntrack_off_fill_match();
> * conntrack_off_fill_tun();
> 
> Is this correct?

I didn't test CT-NAT for IPv6. I just enhanced in a similar way as IPv4 (which I did test)

lib/conntrack.c
```diff
@@ -1373,57 +1373,88 @@ process_one(struct conntrack *ct, struct dp_packet *pkt,
...
+        if (uh->udp_dst != tuple->dst_port) {
+            VLOG_DBG("NAT, change dst port %u from %u",
+                     ntohs(uh->udp_dst), ntohs(tuple->dst_port));
+            item->ct_modify.ipv4.dst_port = uh->udp_dst;
+            item->mod_flags |= CT_OFFLOAD_MODIFY_DST_PORT;
+        }
+    }
+
+    if (reply) {
+        conntrack_swap_tuple_ipv4(&item->ct_modify.ipv4);
```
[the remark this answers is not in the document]
elibritstein: as far as i understood from ophir explanations, as well as the code that was there before (by you).
see commit 82857d0cb8e8 ("conntrack: added offload items")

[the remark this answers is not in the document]
elibritstein: pushed a change. please review

lib/conntrack.c
```diff
@@ -1373,57 +1373,90 @@ process_one(struct conntrack *ct, struct dp_packet *pkt,
...
-                item->mod_flags |= CT_OFFLOAD_MODIFY_SRC_IP;
-            }
-            if (get_16aligned_be32(&l3->ip_dst) != tuple->ipv4_src) {
-                VLOG_DBG("have a nat, chnage dst "IP_FMT" to "IP_FMT, IP_ARGS(
-                           get_16aligned_be32(&l3->ip_dst)),
-                            IP_ARGS(tuple->ipv4_src));
-                item->ct_modify.ipv4.ipv4_dst = tuple->ipv4_src;
-                item->mod_flags |= CT_OFFLOAD_MODIFY_DST_IP;
-            }
+    const struct ip_header *l3 = dp_packet_l3(packet);
```
rbaryanai: This approach might work if you start to track the internals of the CT. if conn has a NAT conn...etc. but this is also a problem because on replay you hit the NAT first. you can track to see how they handling NAT and it is not so simple. this is why I chose to look on what happened to the packet and not on the conntrack internals. see below more specific comment.

lib/conntrack.c
```diff
@@ -1373,57 +1373,90 @@ process_one(struct conntrack *ct, struct dp_packet *pkt,
...
-            }
-            if (get_16aligned_be32(&l3->ip_dst) != tuple->ipv4_dst) {
-                VLOG_DBG("have a nat, chnage dst "IP_FMT" from "IP_FMT,
-                        IP_ARGS( get_16aligned_be32(&l3->ip_dst)),
-                        IP_ARGS(tuple->ipv4_dst));
-                item->ct_modify.ipv4.ipv4_dst =
-                            get_16aligned_be32(&l3->ip_dst);
-                item->mod_flags |= CT_OFFLOAD_MODIFY_DST_IP;
-            }
+    if (get_16aligned_be32(&l3->ip_src) != tuple->ipv4_src) {
```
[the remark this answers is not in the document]
elibritstein: i don't follow. does it mean that the original code was also wrong? let's talk

lib/conntrack.c
```diff
@@ -1373,57 +1373,223 @@ process_one(struct conntrack *ct, struct dp_packet *pkt,
...
+        modify.ipv6_src.s6_addr32[2] = get_16aligned_be32(&l3->ip6_src.be32[2]);
+        modify.ipv6_src.s6_addr32[3] = get_16aligned_be32(&l3->ip6_src.be32[3]);
+        modify.ipv6_dst.s6_addr32[0] = get_16aligned_be32(&l3->ip6_dst.be32[0]);
+        modify.ipv6_dst.s6_addr32[1] = get_16aligned_be32(&l3->ip6_dst.be32[1]);
+        modify.ipv6_dst.s6_addr32[2] = get_16aligned_be32(&l3->ip6_dst.be32[2]);
+        modify.ipv6_dst.s6_addr32[3] = get_16aligned_be32(&l3->ip6_dst.be32[3]);
+        modify.src_port = 0;
+        modify.dst_port = 0;
+    }
+    memcpy(tuple, match_init, sizeof *tuple);
```
rbaryanai: why you copy twice?

lib/conntrack.c
```diff
@@ -1373,57 +1373,223 @@ process_one(struct conntrack *ct, struct dp_packet *pkt,
...
+        }
+        if (uh->udp_dst != tuple->dst_port) {
+            VLOG_DBG("NAT, change dst port %u from %u",
+                     ntohs(uh->udp_dst), ntohs(tuple->dst_port));
+            item->ct_modify.ipv4.dst_port = (modify.dst_port)
+                                            ? tuple->dst_port : uh->udp_dst;
+            item->mod_flags |= CT_OFFLOAD_MODIFY_DST_PORT;
+        }
+    }
+}
```
rbaryanai: looks ok, although complex.

lib/conntrack.c
```diff
@@ -1373,57 +1373,223 @@ process_one(struct conntrack *ct, struct dp_packet *pkt,
...
+        modify.ipv6_dst.s6_addr32[1] = get_16aligned_be32(&l3->ip6_dst.be32[1]);
+        modify.ipv6_dst.s6_addr32[2] = get_16aligned_be32(&l3->ip6_dst.be32[2]);
+        modify.ipv6_dst.s6_addr32[3] = get_16aligned_be32(&l3->ip6_dst.be32[3]);
+        modify.src_port = 0;
+        modify.dst_port = 0;
+    }
+    memcpy(tuple, match_init, sizeof *tuple);
+    if (reply) {
+        conntrack_swap_tuple_ipv6(tuple);
+    }
```
rbaryanai: I think this should be remove, you already swapped.

lib/conntrack.c
elibritstein: two different uses.
with tuple, i might swap it, and it is used to compare.
the modify is taken either from the packet or from the key and it is used to place in the ct->modify

lib/conntrack.c
elibritstein: right. forgot it

PR #65, opened 2019-08-01
Set actions

elibritstein:


lib/netdev-rte-offloads.c
```diff
@@ -3844,6 +3849,29 @@ netdev_rte_offloads_add_set_actions(struct flow_data *fdata,
...
+                fdata->actions.set.ipv4.ttl.ttl_value = key->ipv4_ttl;
+                add_flow_action(flow_actions, RTE_FLOW_ACTION_TYPE_SET_TTL,
+                                &fdata->actions.set.ipv4.ttl);
+            }
+            if (!mask->ipv4_proto || !mask->ipv4_tos || !mask->ipv4_frag) {
+                VLOG_ERR("Unsupported IPv4 set action");
+                return -1;
+            }
         } else {
             VLOG_ERR("Unsupported set action. set_type=%d", set_type);
```
rbaryanai: should be with rate limit. might flood log.

lib/netdev-rte-offloads.c
```diff
@@ -3844,6 +3849,29 @@ netdev_rte_offloads_add_set_actions(struct flow_data *fdata,
                 add_flow_action(flow_actions, RTE_FLOW_ACTION_TYPE_SET_MAC_DST,
                                 &fdata->actions.set.mac.dst);
             }
+        } else if (set_type == OVS_KEY_ATTR_IPV4) {
```
rbaryanai: maybe switch will make it more readable.

lib/netdev-rte-offloads.c
```diff
@@ -3479,11 +3487,11 @@ netdev_dpdk_offload_fill_cls_info(struct netdev_rte_port *rte_port,
...
                     cls_info->actions.pop_tnl = true;
                     cls_info->actions.odp_port = nl_attr_get_odp_port(a);
-                    break;;
-                case OVS_ACTION_ATTR_SET:
-                /*TODO: set baidu eth here. Requirement is not understood */
-
-                break;
+                    break;
+                case OVS_ACTION_ATTR_SET_MASKED:
+                    cls_info->actions.set_actions = nl_attr_get(a);
```
rbaryanai: if flow has more than one? if it has set mac and set ip will both appear on same list?

lib/netdev-rte-offloads.c
```diff
@@ -3511,7 +3519,7 @@ netdev_dpdk_offload_fill_cls_info(struct netdev_rte_port *rte_port,
                 case OVS_ACTION_ATTR_SAMPLE:
                 case OVS_ACTION_ATTR_PUSH_MPLS:
                 case OVS_ACTION_ATTR_POP_MPLS:
-                case OVS_ACTION_ATTR_SET_MASKED:
+                case OVS_ACTION_ATTR_SET:
```
rbaryanai: eth can be called without a mask set.

lib/netdev-rte-offloads.c
```diff
@@ -3479,11 +3487,11 @@ netdev_dpdk_offload_fill_cls_info(struct netdev_rte_port *rte_port,
...
-                    break;;
-                case OVS_ACTION_ATTR_SET:
-                /*TODO: set baidu eth here. Requirement is not understood */
-
-                break;
+                    break;
+                case OVS_ACTION_ATTR_SET_MASKED:
+                    cls_info->actions.set_actions = nl_attr_get(a);
+                    cls_info->actions.set_actions_len = nl_attr_get_size(a);
+                    break;
```
rbaryanai: why not going over actions and make sure we support all, instead of failing late?

lib/netdev-rte-offloads.c
```diff
@@ -3805,19 +3813,66 @@ netdev_dpdk_offload_ct_actions(struct flow_data *fdata,
...
 {
+    int ret;
+
     /* match on vport recirc_id = 0, we must decap first */
     if (cls_info->match.type == MATCH_OFFLOAD_TYPE_VPORT_ROOT) {
         /*TODO: add decap (DONE)*/
         netdev_rte_add_decap_flow_action(flow_actions);
     }
 
+    ret = netdev_rte_offloads_add_set_actions(fdata, flow_actions, cls_info);
```
rbaryanai: can be called also in flow that has CT. set_eth(X), CT(commit), recirc_id(0x9)

lib/netdev-rte-offloads.c
elibritstein: because that way we will have to maintain in different places

lib/netdev-rte-offloads.c
elibritstein: ok

lib/netdev-rte-offloads.c
elibritstein: ok

lib/netdev-rte-offloads.c
elibritstein: how?
i saw in datapath/actions.c, in function execute_set_action:
/* Only tunnel set execution is supported without a mask. */

lib/netdev-rte-offloads.c
elibritstein: yes

elibritstein:
pushed v2

lib/netdev-rte-offloads.c
```diff
@@ -3479,11 +3487,12 @@ netdev_dpdk_offload_fill_cls_info(struct netdev_rte_port *rte_port,
...
                     cls_info->actions.pop_tnl = true;
                     cls_info->actions.odp_port = nl_attr_get_odp_port(a);
-                    break;;
+                    break;
                 case OVS_ACTION_ATTR_SET:
-                /*TODO: set baidu eth here. Requirement is not understood */
-
-                break;
+                case OVS_ACTION_ATTR_SET_MASKED:
+                    cls_info->actions.set_actions = nl_attr_get(a);
```
OphirMunk: In other actions such as CT or CLONE the parsing of nl_attr was done inside this function netdev_dpdk_offload_fill_cls_info(). Why not keeping the same format and do the parsing here instead of just keeping a pointer to nl_attr struct? Our code will look consistent. 
Also, are you sure this pointer will point to valid data once this function is returned?

lib/netdev-rte-offloads.c
```diff
@@ -3755,6 +3763,82 @@ netdev_rte_add_meta_flow_action(struct rte_flow_action_set_meta *meta,
...
+                            cls_info->actions.set_actions_len) {
+        int set_type = nl_attr_type(sa);
+
+        switch ((enum ovs_key_attr) set_type) {
+        case OVS_KEY_ATTR_UNSPEC:
+        case OVS_KEY_ATTR_ENCAP:
+        case OVS_KEY_ATTR_PRIORITY:
+        case OVS_KEY_ATTR_IN_PORT:
+        case OVS_KEY_ATTR_VLAN:
+            VLOG_ERR_RL(&error_rl,
```
OphirMunk: The VLOG should be VLOG_DBG rather then VLOG_ERR. Everything that is not supported will not be offloaded. This is a legitimate flow and there is no error here. Also the message should be something like "not offloaded" rather than "Unsupported".
This comment applies to other appearances of VLOG_ERR as well in this PR.

lib/netdev-rte-offloads.c
OphirMunk: > can be called also in flow that has CT. set_eth(X), CT(commit), recirc_id(0x9)

@elibritstein - what is your reply to this comment?

elibritstein: it was done

lib/netdev-rte-offloads.c
elibritstein: i see it indeed doesn't work with vxlan. debugging.
in this part of netdev_dpdk_offload_fill_cls_info we just check validity
why shouldn't the pointer be valid? it is used in another function but in the same thread in the same code path

lib/netdev-rte-offloads.c
elibritstein: ok

lib/netdev-rte-offloads.c
```diff
@@ -3862,6 +3947,11 @@ netdev_dpdk_offload_put_add_actions(struct netdev_rte_port *rte_port,
 {
     int ret;
 
+    ret = netdev_rte_offloads_add_set_actions(fdata, flow_actions, cls_info);
```
rbaryanai: doesn't seems right. sometime you need to first decap, how it should work? you will first change the vxlan mac/ip and then decap.

lib/netdev-rte-offloads.c
```diff
@@ -3806,7 +3811,31 @@ netdev_rte_offloads_add_set_actions(struct flow_data *fdata,
             }
             } break;
         case OVS_KEY_ATTR_ETHERTYPE:
```
rbaryanai: I think there is a missing break here, or error.

lib/netdev-rte-offloads.c
elibritstein: fixed

lib/netdev-rte-offloads.c
rbaryanai: what about the missing break?

lib/netdev-rte-offloads.c
```diff
@@ -3755,6 +3768,106 @@ netdev_rte_add_meta_flow_action(struct rte_flow_action_set_meta *meta,
...
+
+    NL_ATTR_FOR_EACH_UNSAFE(sa, sleft, cls_info->actions.set_actions,
+                            cls_info->actions.set_actions_len) {
+        int set_type = nl_attr_type(sa);
+
+        switch ((enum ovs_key_attr) set_type) {
+        case OVS_KEY_ATTR_UNSPEC:
+        case OVS_KEY_ATTR_ENCAP:
+        case OVS_KEY_ATTR_PRIORITY:
+        case OVS_KEY_ATTR_IN_PORT:
```
rbaryanai: break

elibritstein: there is a break few lines after here
and regarding the above, no missing break

PR #66, opened 2019-08-01
netdev-rte-offloads: Add match on no-fragmentation

elibritstein:
Issue: 1844140
Signed-off-by: Eli Britstein <elibr@mellanox.com>

don't merge until the following is resolved:

[MLNX OFED - Bug SW #1846492] (Rejected) cannot set fragment_offset mask From: Asaf Penso created on 01/08/2019 09:30:03 This feature is not supported. Please talk to Marketing (Erez Scop) and create a new feature request. We'll handle per priorities.

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -510,7 +510,7 @@ add_flow_rss_action(struct flow_actions *actions,
 }
 
 enum {
-    REG_IDX_RECIRC_ID = 0,
+    REG_IDX_HW_ID = 0,
     REG_IDX_CT_ZONE,
     REG_IDX_CT_MARK,
     REG_IDX_OUTER_ID,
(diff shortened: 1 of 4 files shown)


lib/netdev-rte-offloads.c
```diff
@@ -613,6 +613,8 @@ add_flow_patterns(struct flow_patterns *patterns,
         mask->ipv4.hdr.next_proto_id   = match->wc.masks.nw_proto;
         mask->ipv4.hdr.src_addr        = match->wc.masks.nw_src;
         mask->ipv4.hdr.dst_addr        = match->wc.masks.nw_dst;
+        mask->ipv4.hdr.fragment_offset = RTE_IPV4_HDR_OFFSET_MASK |
+                                         RTE_IPV4_HDR_MF_FLAG;
```
OphirMunk: You assign: fragment_offset = 16383
Can you please explain?

elibritstein: the "fragment_offset" field in DPDK is 16 bit, meaning it has both "flags" and "fragment_offset" fields of IPv4 header (see https://en.wikipedia.org/wiki/IPv4)
the lower 13 bits are the fragment offset, and the 14th bit is bit 2 of flags: More Fragments (MF).
bitwise-or of 2^13-1 and 1<<14 is 0x3fff, which is this value.
notice this is the mask.
the value (spec) is zero.
this means we match those bits are 0

PR #70, opened 2019-08-05
dpif-netdev: Fix race condition during MTU update.

OphirMunk:
This commits fixes a race condition which occurs under the following
conditions in OVS-DPDK:
1. OVS is configured with hw-offload=true.
2. Some offloaded traffic (e.g. ping packets) continuously pass through OVS
3. A phy port MTU size is updated:
ovs-vsctl set Interface <PHY PORT NAME> mtu_request=<NEW MTU>

As a result a segmentation fault may occur, as logged in [1].  During an
MTU update the datapath is reconfigured which may lead to the following
race condition:
1. flow_mark_flush()is called. The sender thread walks through all entries
in mark_to_flow CMAP and sends a request per flow to
"dp_netdev_flow_offload" thread to delete the individual entires from the
CMAP.
2. flow_mark_flush() is called again before any entry was removed and the
same set of requests is sent.
3. The "dp_netdev_flow_offload" thread now has two identical series
of removal requests. The first series will succeed while the second series
will fail as all entries have already been removed from the CMAP.
This commit verifies that a CMAP entry exists before automatically removing
it, otherwise it ignores the request.

[1] lib/cmap.c:922: assertion cmap_replace__(impl, old_node, new_node,
hash, h1) || cmap_replace__(impl, old_node, new_node, hash, h2) failed i n
cmap_replace()

Fixes: 241bad15d99a ("dpif-netdev: associate flow with a mark id")


lib/dpif-netdev.c
```diff
@@ -2401,6 +2401,23 @@ flow_mark_has_no_ref(uint32_t mark)
...
+static struct dp_netdev_flow *
+mark_to_flow_find(const struct dp_netdev_pmd_thread *pmd,
+                  const uint32_t mark)
+{
+    struct dp_netdev_flow *flow;
+
+    CMAP_FOR_EACH_WITH_HASH (flow, mark_node, hash_int(mark, 0),
+                             &flow_mark.mark_to_flow) {
+        if (flow->mark == mark && flow->pmd_id == pmd->core_id &&
+            flow->dead == false) {
```
rbaryanai: This will fail you even if the flow exits, because del flow will set flow to dead mode.

lib/dpif-netdev.c
```diff
@@ -2410,6 +2427,11 @@ mark_to_flow_disassociate(struct dp_netdev_pmd_thread *pmd,
     struct cmap_node *mark_node = CONST_CAST(struct cmap_node *,
                                              &flow->mark_node);
 
+    struct dp_netdev_flow *f = mark_to_flow_find(pmd, mark);
```
rbaryanai: seems wrong approach to me. just check if flow->mark != INVALID_FLOW_MARK. beside the dead issue described above, you might be removing the wrong flow, because this flow might have been already removed and another flow taken its mark (although couldn't find a real scenario easily)

lib/dpif-netdev.c
OphirMunk: This is legacy code moved up a few lines. This code has survived all our past tests. Not sure we have issues here.

rbaryanai: I know, but it means that any flow->dead = true will not be found and you will not return the mark

lib/dpif-netdev.c
OphirMunk: I disagree since this code correctly access a CMAP table. Basing it on application criteria mark != INVALID_FLOW_MARK seems wrong. 
Let's discuss your concern regarding the dead flag.

rbaryanai: seems like you will hide a bug here.

OphirMunk:
@rbaryanai - can you please review?

lib/dpif-netdev.c
```diff
@@ -7215,12 +7219,12 @@ dp_netdev_input__(struct dp_netdev_pmd_thread *pmd,
 
     /* All the flow batches need to be reset before any call to
      * packet_batch_per_flow_execute() as it could potentially trigger
-     * recirculation. When a packet matching flow ‘j’ happens to be
+     * recirculation. When a packet matching flow ???j??? happens to be
```
rbaryanai: what is this?

OphirMunk: Typo. Fixed now.

OphirMunk:
@rbaryanai - please review now

PR #71, opened 2019-08-06
Ct zone

elibritstein:
i see those commits reveal a bug - a crash in stopping the OVS.
the crash is in netdev_dpdk_get_hw_id.
keeping the OVS up, i don't see a problem (probably hidden).
i issue this PR anyway, for review and for Baidu

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -511,11 +511,43 @@ add_flow_rss_action(struct flow_actions *actions,
 }
 
 enum {
-    REG_IDX_HW_ID = 0,
-    REG_IDX_CT_MARK,
-    REG_IDX_OUTER_ID,
-    REG_IDX_CT_STATE,
-    REG_IDX_NUM,
+    TAG_FIELD_HW_ID = 0,
+    TAG_FIELD_CT_STATE,
+    TAG_FIELD_CT_ZONE,
+    TAG_FIELD_OUTER_ID,
+    TAG_FIELD_CT_MARK,
+    TAG_FIELD_NUM,
+};
+
+/* map tag_id to register index, mask as few tags can share the same register
+ * and data field is used as the bit shift needed.
+ */
+static struct rte_flow_action_set_tag tag_id_maps[] = {
+    [TAG_FIELD_HW_ID] = {
+        .index = 0,
+        .mask = 0xFFFFFFFF,
+        .data = 0,
+    },
+    [TAG_FIELD_CT_STATE] = {
+        .index = 1,
+        .mask = 0x000000FF,
+        .data = 0,
+    },
+    [TAG_FIELD_CT_ZONE] = {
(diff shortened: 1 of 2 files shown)


lib/netdev-rte-offloads.c
```diff
@@ -511,11 +511,43 @@ add_flow_rss_action(struct flow_actions *actions,
...
+    TAG_FIELD_NUM,
+};
+
+/* map tag_id to register index, mask as few tags can share the same register
+ * and data field is used as the bit shift needed.
+ */
+static struct rte_flow_action_set_tag tag_id_maps[] = {
+    [TAG_FIELD_HW_ID] = {
+        .index = 0,
+        .mask = 0xFFFFFFFF,
```
rbaryanai: HW id should be 16 bits

lib/netdev-rte-offloads.c
```diff
@@ -511,11 +511,43 @@ add_flow_rss_action(struct flow_actions *actions,
...
+        .data = 0,
+    },
+    [TAG_FIELD_CT_STATE] = {
+        .index = 1,
+        .mask = 0x000000FF,
+        .data = 1,
```
rbaryanai: should not it be zero shift?

lib/netdev-rte-offloads.c
```diff
@@ -511,11 +511,43 @@ add_flow_rss_action(struct flow_actions *actions,
...
+    TAG_FIELD_CT_MARK,
+    TAG_FIELD_NUM,
+};
+
+/* map tag_id to register index, mask as few tags can share the same register
+ * and data field is used as the bit shift needed.
+ */
+static struct rte_flow_action_set_tag tag_id_maps[] = {
+    [TAG_FIELD_HW_ID] = {
+        .index = 0,
```
rbaryanai: index 0/1 are reserved for meta and mark and shouldn't be used. I know this is a PMD issue but do they handle it?

lib/netdev-rte-offloads.c
elibritstein: hw id should not be in a tag at all. will be in metadata (in mlx case 16 bit)

lib/netdev-rte-offloads.c
elibritstein: it is different API in DPDK. the app doesn't know about it, and can use index 0/1 too

rbaryanai: great

lib/netdev-rte-offloads.c
rbaryanai: So why is it here?

elibritstein: because it is not the in the scope of those commits

lib/netdev-rte-offloads.c
elibritstein: fixed

PR #76, opened 2019-08-08
Ct miss handling

OphirMunk:


lib/netdev-rte-offloads.c
```diff
@@ -4305,14 +4352,31 @@ netdev_dpdk_offload_put_handle(struct netdev *netdev,
...
-    /* for all cases, we need to save all resources allocated */
-    ret = netdev_dpdk_save_flow_miss_ctx(info->flow_mark,
-                                         cls_info.actions.hw_id,
-                                         !cls_info.actions.recirc_id,
-                                         cls_info.match.outer_id,
-                                         match->flow.in_port.odp_port,
-                                         cls_info.actions.type == ACTION_OFFLOAD_TYPE_CT);
-    if (ret) {
+
+    int miss_type;
```
rbaryanai: I don't understand the logic here. you have 3 types of miss:
1. first is miss on vxlan (or tun port) table, in this case mark == VXLAN_TABLE_ID. this has no flow connected to it (like was already implemented).
2. miss on flow with CT - this means we get the flow mark and we need to restore tun if exists, and we need to cut the actins (like Oz did, but for all flows).
3. other miss - on other miss you get the recirc_id (meta_data) and you get the CT session mark. in this case you need to store the zone,mark,label...etc and tun info (outer) if exists. fix port_id if needed.

lib/netdev-rte-offloads.c
```diff
@@ -4305,14 +4352,31 @@ netdev_dpdk_offload_put_handle(struct netdev *netdev,
...
+    switch (cls_info.actions.type) {
+    case ACTION_OFFLOAD_TYPE_CT:
+        miss_type = MARK_PREPROCESS_FLOW_WITH_CT;
```
rbaryanai: OK. you relace single CT or not with type.

lib/netdev-rte-offloads.c
```diff
@@ -4305,14 +4352,31 @@ netdev_dpdk_offload_put_handle(struct netdev *netdev,
...
+        break;
+    case ACTION_OFFLOAD_TYPE_OUTPUT:
+    case ACTION_OFFLOAD_TYPE_TNL_PUSH:
+        miss_type = MARK_PREPROCESS_FLOW;
```
rbaryanai: not a use case. those flows if you hit them packet is going out, so no need.

lib/netdev-rte-offloads.c
```diff
@@ -4305,14 +4352,31 @@ netdev_dpdk_offload_put_handle(struct netdev *netdev,
...
+        break;
+    case ACTION_OFFLOAD_TYPE_TNL_POP:
+        miss_type = MARK_PREPROCESS_VXLAN;
```
rbaryanai: you never see it, because if you hit the flow you jump to the vxlan port and only there you can miss, and in this case we already know what to do. we have a default flow with low priority that set mark to the special mark.

lib/netdev-rte-offloads.c
```diff
@@ -4469,6 +4534,7 @@ netdev_dpdk_ct_ctx_get_ref_outer_id(struct mark_to_miss_ctx_data *data,
     }
     if (ct_offload2->tun.ip_dst) {
         data->ct.outer_id[dir2] = netdev_dpdk_tun_id_get_ref(
+                                       ct_offload2->odp_port,
```
rbaryanai: seems unrelated. why you keep the odp_port in the tun info, we already have a generic data sturcture which is called miss_ctx

lib/netdev-rte-offloads.c
```diff
@@ -5005,38 +5072,67 @@ restore_packet_state(uint32_t flow_mark, struct dp_packet *packet)
...
-    else if ((packet->md.in_port.odp_port == ct_rep->odp_port) ||
-            (rte_port->is_uplink && ct_rep->tun.ip_dst)) {
-        ct_offload = ct_rep;
-    } else {
-        VLOG_DBG("Could not match any direction for CT miss handling");
-    }
+    switch (miss_ctx->type) {
+    case MARK_PREPROCESS_VXLAN:
+        /* Pop header that will also restore meta data */
+        packet = netdev_vxlan_pop_header(packet);
```
rbaryanai: calling function is not checking returned value, you get crash if pop fails. but as I see it you will never get to this point anyway.

lib/netdev-rte-offloads.c
```diff
@@ -5005,38 +5072,67 @@ restore_packet_state(uint32_t flow_mark, struct dp_packet *packet)
...
-    }
-    if (hw_id && netdev_dpdk_get_id_from_hw_id(hw_id, false,
-                                               &packet->md.recirc_id)) {
-        VLOG_DBG("Failed to get recirc_id from hw_id %u", hw_id);
+    case MARK_PREPROCESS_CT:
+        ct_init = miss_ctx->ct.ct_offload[CT_OFFLOAD_DIR_INIT];
+        ct_rep = miss_ctx->ct.ct_offload[CT_OFFLOAD_DIR_REP];
+        rte_port = netdev_rte_port_search(packet->md.in_port.odp_port, &port_map);
+        if ((packet->md.in_port.odp_port == ct_init->odp_port) ||
+           (rte_port->is_uplink && ct_init->tun.ip_dst)) {
```
rbaryanai: again, why not saving it the ctx and not on the outer.

lib/netdev-rte-offloads.c
OphirMunk: @rbaryanai  - thanks for your review. Updated the PR based on it.

OphirMunk:
@ozshlomo  - please review latest fix for double referencing hw id.

lib/netdev-rte-offloads.c
```diff
@@ -4441,9 +4441,7 @@ netdev_dpdk_offload_put_handle(struct netdev *netdev,
                                        cls_info.match.tun_id);
     }
 
-    if (cls_info.match.hw_id != INVALID_HW_ID) {
-        netdev_dpdk_put_recirc_id_hw_id(cls_info.match.hw_id);
-    }
+    /* No need to put the hw_id as it is done when the flow is deleted */
```
ozshlomo: This does not fix the problem.

The issue is that there is a put without a get.
When rolling back in my scenario the hw_id is still INVALID at this point

ozshlomo: Please revert this commit in the PR

lib/netdev-rte-offloads.c
```diff
@@ -1846,7 +1846,7 @@ ct_add_rte_flow_offload(struct netdev_rte_port *rte_port,
 }
 
 static struct mark_to_miss_ctx_data *
-    netdev_dpdk_get_flow_miss_ctx(uint32_t mark);
+    netdev_dpdk_get_flow_miss_ctx(uint32_t mark, bool create);
```
elibritstein: we have netdev_dpdk_find_miss_ctx
why do you need this?

lib/netdev-rte-offloads.c
```diff
@@ -5082,12 +5082,10 @@ restore_packet_state(uint32_t flow_mark, struct dp_packet *packet)
 
     switch (miss_ctx->type) {
     case MARK_PREPROCESS_VXLAN:
-        /* Pop header that will also restore meta data */
-        packet = netdev_vxlan_pop_header(packet);
-        if (!packet) {
-            return -1;
-        }
-        return 0;
+        /*
```
elibritstein: if so, why do we need MARK_PREPROCESS_VXLAN for?

OphirMunk:
@elibritstein - thanks for your review. As a decision was taken to merge this PR as is I will implement the fixes in new commits.

PR #77, opened 2019-08-08
Ct nat ol fixes

elibritstein:


lib/netdev-rte-offloads.c
```diff
@@ -833,6 +833,10 @@ add_flow_patterns(struct flow_patterns *patterns,
         break;
     }
 
+    if (match->flow.recirc_id == 0) {
+        return 0;
+    }
+
```
rbaryanai: is it possible unless ct_state= 0?

elibritstein: yes. recirc is not only for ct.
there may be other rules that involve recirc

elibritstein: it is not possible if ct_state != 0, as ct_state uses recirc mechanism.
i meant the correct check for that is recirc, and not some metadata field (ct_state) that uses it.

rbaryanai: but is mask exists, doesn't it means that we are doing wrong match here?
if recirc_id = 0, and ct_state has a value and ct_state mask is not zero then we compare wrong.

rbaryanai: sorry, I don't get what is the purpose if the "if", it is covered anyway later because the mask will be zero.

elibritstein: example:
ufid:cd970008-ef86-4a1b-8c71-089afe3dc9f8, skb_priority(0/0),skb_mark(0/0),ct_state(0/0x20),ct_zone(0/0),ct_mark(0/0),ct_label(0/0),recirc_id(0),dp_hash(0/0),in_port(vm1),packet_type(ns=0,id=0),eth(src=e4:11:22:33:44:50/00:00:00:00:00:00,dst=e4:11:22:33:44:51/00:00:00:00:00:00),eth_type(0x0800),ipv4(src=4.4.4.10/0.0.0.0,dst=4.4.4.11/0.0.0.0,proto=6,tos=0/0,ttl=64/0,frag=no),tcp(src=38380/0,dst=5001/0),tcp_flags(0/0), packets:13715278, bytes:20763626244, used:0.431s, flags:., offloaded:yes, dp:ovs-dpdk, actions:ct(zone=2),recirc(0x3)

here, ct_state(0/0x20). it means that without that if we'll match that, but i just think it's not needed.

rbaryanai: Ok, maybe you need assert on ct_state != 0 in this case.

elibritstein: what for?

PR #78, opened 2019-08-11
Free miss ctx

elibritstein:


lib/netdev-rte-offloads.c
```diff
@@ -4653,9 +4653,20 @@ netdev_dpdk_offload_ct_put(struct ct_flow_offload_item *ct_offload,
     if (data->ct.ct_offload[dir_opp]) {
         struct ct_flow_offload_item *ct_off_opp = data->ct.ct_offload[dir_opp];
 
-        data->ct.ct_offload[dir_opp]->ct_state = ct_offload->ct_state;
-        if (netdev_dpdk_offload_ct_session(data, ct_off_opp, ct_offload)) {
+        if (ct_offload->ct_state & CS_ESTABLISHED &&
```
ozshlomo: Please review this logic.
There should not be any assumptions on the ct_state. Nor should we change the ct_state flags

PR #80, opened 2019-08-13
netdev-dpdk: Enable scatter offload regardless of MTU value

evelineraine:
Initially scatter was enabled only to support MTUs larger than the
standard Ethernet MTU of 1500. In order to support small MTUs (< 1000)
scatter is also required as the mbuf size is too small to fit the whole
frame.

------- 

When MTU on PF is set to a value lower than 999, OVS allocates 1152 bytes of mbuf, which is not enough to fit a 1500 bytes frame. Startup fails with the following log messages:
```
ovs|00047|dpdk|ERR|net_mlx5: port 0 Rx queue 0: Scatter offload is not configured and no enough mbuf space(1152) to contain the maximum RX packet length(1518) with head-room(128)
ovs|00048|dpdk|ERR|net_mlx5: port 0 unable to allocate queue index 0
ovs|00050|netdev_dpdk|ERR|Interface pf(rxq:1 txq:2 lsc interrupt mode:false) configure error: Cannot allocate memory
ovs|00051|dpif_netdev|ERR|Failed to set interface pf new configuration
```

The obvious fix is to enable RX scatter offload (which appears to be off by default for mlx5 pmd) for any MTU size (before it was enabled for jumbo frames only).

Please advise if this is a sane thing to do.

*Note*: MTUs below 1000 are an option only for IPv4, IPv6 allows minimum MTUs of 1280 bytes, so the issue would not reproduce.

### DIFF
--- a/lib/netdev-dpdk.c
+++ b/lib/netdev-dpdk.c
@@ -919,13 +919,11 @@ dpdk_eth_dev_port_config(struct netdev_dpdk *dev, int n_rxq, int n_txq)
     rte_eth_dev_info_get(dev->port_id, &info);
 
     /* As of DPDK 17.11.1 a few PMDs require to explicitly enable
-     * scatter to support jumbo RX.
+     * scatter to support jumbo RX or MTUs below 1k.
      * Setting scatter for the device is done after checking for
      * scatter support in the device capabilites. */
-    if (dev->mtu > RTE_ETHER_MTU) {
-        if (dev->hw_ol_features & NETDEV_RX_HW_SCATTER) {
-            conf.rxmode.offloads |= DEV_RX_OFFLOAD_SCATTER;
-        }
+    if (dev->hw_ol_features & NETDEV_RX_HW_SCATTER) {
+        conf.rxmode.offloads |= DEV_RX_OFFLOAD_SCATTER;
     }
 
     conf.intr_conf.lsc = dev->lsc_interrupt_mode;


lib/netdev-dpdk.c
```diff
@@ -919,13 +919,11 @@ dpdk_eth_dev_port_config(struct netdev_dpdk *dev, int n_rxq, int n_txq)
...
     /* As of DPDK 17.11.1 a few PMDs require to explicitly enable
-     * scatter to support jumbo RX.
+     * scatter to support jumbo RX or MTUs below 1k.
      * Setting scatter for the device is done after checking for
      * scatter support in the device capabilites. */
-    if (dev->mtu > RTE_ETHER_MTU) {
-        if (dev->hw_ol_features & NETDEV_RX_HW_SCATTER) {
-            conf.rxmode.offloads |= DEV_RX_OFFLOAD_SCATTER;
-        }
+    if (dev->hw_ol_features & NETDEV_RX_HW_SCATTER) {
```
elibritstein: i think it's better to fix the condition to something involves MTU_TO_FRAME_LEN and/or FRAME_TO_MTU_LEN macros.
the purpose is to enable it only when it is really needed, i.e. when the device's specific mtu is larger than the allocated space per mbuf.

evelineraine: What is `device's specific mtu` in this case? 
Supposedly we should set MTU to 998, but still DPDK sanity checks the mbuf size against MTU of 1500. Why would receiving a 1.5k frame be a possibility in this case?

elibritstein: i don't fully understand the mempool model of ovs. in other applications i saw there was one mempool (or one per numa socket), here, in ovs, i'm not sure.
if there can be multiple mempools, each with different element size, need to understand how ovs handles this.
if the element size of all mempools is the same, this is the one that matters.
now, we need to compare the device's mtu (i suppose dev->mtu), to the element size of the mempool.
if dev->mtu is larger, so we need to use the scatter feature if available. if this feature is not available, we should fail the configuration

evelineraine: Ok, I will look how OVS choses the size of mempool (it's probably based on MTU, but it's not proportional, looks like it decreases size in steps - mbuf size of 1152 is set only for MTUs lower than 999).
But my question is about `dev->mtu` value. OVS does configure the device to use the specified MTU from `mtu_request=<X>` - it can even be seen with `ip link` - the PF MTU is changed. But when MTU is 998, DPDK should not worry to receive a 1500 bytes frame.

elibritstein: i don't understand your question

PR #83, opened 2019-08-20
netdev-rte-offloads: Adjust encap stats

OphirMunk:
When running with hw-offload=true the HW bytes counter of flows with
encap action includes the encap header size. In order to adjust the
counter to OVS operation with hw-offload=false this extra bytes count
should be subtracted.


lib/netdev-rte-offloads.c
```diff
@@ -1336,7 +1341,7 @@ netdev_rte_offloads_flow_stats_get(struct netdev *netdev OVS_UNUSED,
                     continue;
                 }
                 stats->n_packets += (query.hits_set) ? query.hits : 0;
-                stats->n_bytes += (query.bytes_set) ? query.bytes : 0;
+                stats->n_bytes += (query.bytes_set) ? (query.bytes + (stats->n_packets * uho->hw_stats_adjust)) : 0;
```
elibritstein: no need for extra ()

lib/netdev-rte-offloads.c
```diff
@@ -3924,18 +3929,19 @@ netdev_dpdk_offload_tnl_push(struct flow_data *fdata,
                              struct flow_actions *flow_actions,
                              struct offload_item_cls_info *cls_info)
 {
+    /* TODO: count (Done) */
```
elibritstein: if it doesn't matter the order, why change?

lib/netdev-rte-offloads.c
```diff
@@ -3924,18 +3929,19 @@ netdev_dpdk_offload_tnl_push(struct flow_data *fdata,
...
+
     /* TODO: encap (Done) */
     fdata->actions.clone_raw_encap.data =
                              (uint8_t *)cls_info->actions.push_tnl->header;
     fdata->actions.clone_raw_encap.preserve = NULL;
-    fdata->actions.clone_raw_encap.size =
-                                       cls_info->actions.push_tnl->header_len;
+    uint32_t header_len = cls_info->actions.push_tnl->header_len;
+    fdata->actions.clone_raw_encap.size = header_len;
+    fdata->hw_stats_adjust = -header_len;
```
elibritstein: would be clearer to have a positive number, and reduce in the calculation

OphirMunk: I prefer it this way. Here exactly you know the adjustment is negative. Remark: maybe better to use '-=' rather than '='. In other places we may have positive adjustment. So in the calculation we just adjust.

lib/netdev-rte-offloads.c
OphirMunk: We write it in SW as we expect HW to behave (not to reorder actions) so once and if the HW is updated we are already good is SW.

lib/netdev-rte-offloads.c
OphirMunk: OK

lib/netdev-rte-offloads.c
elibritstein: the expected behavior (not implemented yet in rdma-core) is that if we put the count before, it won't count the encap header, and if we put it after, it would.
so, in this workaround, to work even if they implement it, we need to have the count after the encap

OphirMunk: OK

PR #84, opened 2019-08-22
Misc

elibritstein:
- minor cosmetic commits
- in a pre-step to ct-stas, fixed a bug in removing ct from port map
- fix missing ct-zone match in ct-table

### DIFF
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -3427,6 +3427,7 @@ netdev_dpdk_del_miss_ctx(uint32_t mark)
     CMAP_FOR_EACH_WITH_HASH (data, node, hash,
                       &mark_to_miss_ctx) {
         if (data->mark == mark) {
+                VLOG_DBG("Free miss context for mark=%d", mark);
                 cmap_remove(&mark_to_miss_ctx,
                         CONST_CAST(struct cmap_node *, &data->node), hash);
                 ovsrcu_postpone(free, data);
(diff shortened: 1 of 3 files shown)


lib/netdev-rte-offloads.c
```diff
@@ -1376,7 +1403,7 @@ netdev_offloads_flow_del(const ovs_u128 *ufid, struct cmap *cmap,
 
             if (netd == rte_port->netdev && hw_offload->rte_flow_data[i].flow != NULL) {
                 struct ufid_to_odp *uto =
-                    ufid_to_portid_get(ufid, &ufid_to_portid_map);
+                    ufid_to_portid_get(ufid, cmap);
```
elibritstein: this function is used for ct del as well. this is a bug fix

lib/netdev-rte-offloads.c
```diff
@@ -4632,6 +4632,7 @@ fill_ct_match(struct match *match, struct ct_flow_offload_item *item)
     match->wc.masks.tp_dst = 0xFFFF;
     match->flow.ct_zone = item->zone;
     match->wc.masks.ct_zone = 0xFFFF;
+    match->flow.recirc_id = 1;
```
OphirMunk: 1. This requires a comment in the code (can copy from commit message).
```
/*
 * Metadata matches are not done in recirc 0. In CT matches, 
 * set the recirc to non-zero value.
 */
```
2. Better use a macro rather than 1. For example:
```
#define NON_ZERO_RECIRC_ID 0xffffffff

match->flow.recirc_id = NON_ZERO_RECIRC_ID;
```

PR #85, opened 2019-08-22
netdev-dpdk: Protect rte_flow_query call with mutex

OphirMunk:
All dpdk APIs (e.g rte_flow_create, rte_flow_destroy) must be mutex
protected. This commit protects rte_flow_query API.

### DIFF
--- a/lib/netdev-dpdk.c
+++ b/lib/netdev-dpdk.c
@@ -4856,6 +4856,22 @@ netdev_dpdk_rte_flow_create(struct netdev *netdev,
     return flow;
 }
 
+int
+netdev_dpdk_rte_flow_query(struct netdev *netdev,
+                           struct rte_flow *flow,
+                           const struct rte_flow_action *action,
+                           void *data,
+                           struct rte_flow_error *error)
+{
+    int ret;
+    struct netdev_dpdk *dev = netdev_dpdk_cast(netdev);
+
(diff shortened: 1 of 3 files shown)


lib/netdev-dpdk.c
```diff
@@ -4856,6 +4856,22 @@ netdev_dpdk_rte_flow_create(struct netdev *netdev,
...
+{
+    int ret;
+    struct netdev_dpdk *dev = netdev_dpdk_cast(netdev);
+
+    ovs_mutex_lock(&dev->mutex);
+    ret = ovs_rte_flow_query(dev->port_id, flow, action, data, error);
+    ovs_mutex_unlock(&dev->mutex);
+    return ret;
+}
+
```
ozshlomo: Ophir, please check why a query action needs a mutex.
Where are the potential race conditions?

rbaryanai: the statistics calls are called from different thread and basically it is protected because we take port_mutex on DP, which is also taken on put/del flow and also if there is a DP reconfiguration. however, this was discussed in the past, and it was decided for now, that we take the dev->mutex on flow/put and since get_stats is also api called (and it is not protected by design by reading currently destroyed flow...etc). mutex should be taken. 
the reason that community (ilya), think that the offload way might change and at some point we no longer guarantee that  dp_mutex is used.

PR #86, opened 2019-08-27
Ct aging

OphirMunk:


lib/conntrack.c
```diff
@@ -1799,6 +1799,18 @@ set_label(struct dp_packet *pkt, struct conn *conn,
...
+static bool
+conn_hw_active(struct conntrack *ct,
+               struct conn *conn)
+{
+    struct ct_flow_offload_item off_item;
+
+    memset(&off_item, 0, sizeof off_item);
+    conntrack_off_fill_key(&off_item, &conn->key);
+    return netdev_is_flow_api_enabled() &&
+           ct && ct->off_class && ct->off_class->conn_active(&off_item);
```
elibritstein: check conn_active is not NULL

lib/dpif-netdev.c
```diff
@@ -2699,6 +2699,7 @@ static void ct_offload_dump_ipv6_item(struct ds *s,
 static void ct_offload_dump_item(struct ds *s,
                                  struct ct_flow_offload_item *item)
 {
+    ds_put_format(s, "    IPv%d, zone=%d\n", item->ct_ipv6 ? 6 : 4, item->zone);
```
elibritstein: this belongs to another commit

lib/conntrack.c
```diff
@@ -1701,10 +1708,12 @@ conntrack_add_ct_off_item(struct conntrack *ct,
         }
 
         if (ctx->reply && !(ctx->conn->offload_flags & CT_OFF_REP)) {
+                (void)&get_tm;
```
elibritstein: i think those changes in lib/conntrack.c can be in another commit, that needs them instead of this dummy

lib/conntrack-icmp.c
```diff
@@ -45,6 +45,12 @@ conn_icmp_cast(const struct conn *conn)
     return CONTAINER_OF(conn, struct conn_icmp, up);
 }
 
+static enum ct_timeout
+icmp_get_tm(struct conn *conn_ OVS_UNUSED)
+{
+    return 0;
```
elibritstein: don't use 0. use a proper value from enum ct_timeout

lib/conntrack-private.h
```diff
@@ -107,6 +107,7 @@ struct conn {
     /* True if alg data connection. */
     uint8_t alg_related;
     uint8_t offload_flags;
+    enum ct_timeout tm;
```
elibritstein: there is no need to keep it. we can just call the function when it's needed

lib/conntrack.c
```diff
@@ -1799,6 +1799,18 @@ set_label(struct dp_packet *pkt, struct conn *conn,
     }
 }
 
+static bool
```
elibritstein: should be squashed with the commit using it

lib/netdev-rte-offloads.c
```diff
@@ -4845,8 +4853,29 @@ netdev_dpdk_offload_ct_del(uint32_t mark)
...
+    if (!netdev_dpdk_find_miss_ctx(mark, &data)) {
+        return false;
+    }
+
+    for (active = false, dir = CT_OFFLOAD_DIR_INIT;
+         dir < CT_OFFLOAD_NUM && !active; dir++) {
+        build_ctid(data->mark, dir, false, &ctid);
+        memset(&stats, 0, sizeof stats);
+        if (net_rte_offloads_ct_stats_get(&ctid, &stats)) {
+            VLOG_WARN("%s failed for mark=%d", __FUNCTION__, mark);
```
elibritstein: diff --git a/lib/netdev-rte-offloads.c b/lib/netdev-rte-offloads.c
index 3288fd618..a7ed3da21 100755
--- a/lib/netdev-rte-offloads.c
+++ b/lib/netdev-rte-offloads.c
@@ -4850,6 +4850,12 @@ netdev_dpdk_offload_ct_active(uint32_t mark)
     int dir;
 
     if (!netdev_dpdk_find_miss_ctx(mark, &data)) {
+        VLOG_WARN("%s: cannot find context for mark=%d", __FUNCTION__, mark);
+        return false;
+    }
+    if (!(data->ct.ct_offload[CT_OFFLOAD_DIR_INIT]->ct_state & CS_ESTABLISHED) ||
+        !(data->ct.ct_offload[CT_OFFLOAD_DIR_REP]->ct_state & CS_ESTABLISHED)) {
+        VLOG_DBG("%s: ct for mark=%d is not established", __FUNCTION__, mark);

lib/conntrack.c
OphirMunk: OK.
Please note this code is part of a static function conn_hw_active() which is unused. 
It causes a compilation warning which will have to be resolved for upstream.

elibritstein: i commented that it should be squashed with the commit using it

lib/dpif-netdev.c
OphirMunk: This line was  added in this commit 21bfb54aa5b9 ("conntrack: Add API for connection active in HW") while the containing function was added in commit 4a8fd4a08296 ("netdev-rte-offloads: ct-put and ct-del events") which is already part of ct-integration branch.
Nothing to do unless we decide to overwrite (push force) ct-integration branch.
In any case it must be sorted out for upstream.

elibritstein: i had another commit for it in my PR
3b3951f46 conntrack: Dump IP ver and zone
i don't know why you squashed it with this commit

lib/netdev-rte-offloads.c
OphirMunk: Added

lib/conntrack.c
OphirMunk: Done

lib/conntrack-private.h
OphirMunk: Done

lib/conntrack-icmp.c
OphirMunk: Done

lib/conntrack.c
OphirMunk: Removed

lib/dpif-netdev.c
OphirMunk: Fixed

OphirMunk:
@elibritstein - please review updated PR 86

lib/conntrack-other.c
```diff
@@ -42,6 +42,12 @@ conn_other_cast(const struct conn *conn)
     return CONTAINER_OF(conn, struct conn_other, up);
 }
 
+static int
+other_get_tm(struct conn *conn_ OVS_UNUSED, enum ct_timeout *tc OVS_UNUSED)
```
elibritstein: typo (tc instead of tm)

lib/conntrack.c
elibritstein: not removed

lib/conntrack.c
```diff
@@ -2939,6 +2962,13 @@ conntrack_dump_next(struct conntrack_dump *dump, struct ct_dpif_entry *entry)
             INIT_CONTAINER(conn, node, node);
             if ((!dump->filter_zone || conn->key.zone == dump->zone) &&
                  (conn->conn_type != CT_CONN_TYPE_UN_NAT)) {
+                if (conn_hw_active(ct, conn)) {
```
elibritstein: please recheck this location, without dump-conntrack in the background

lib/conntrack.c
```diff
@@ -2939,6 +2962,13 @@ conntrack_dump_next(struct conntrack_dump *dump, struct ct_dpif_entry *entry)
...
+                    enum ct_timeout tm;
+                    if (!get_tm(conn, &tm)) {
```
elibritstein: i suggest to call get_tm inside conn_hw_active.
there is no point to check the HW rules if we don't have a tm to update

lib/conntrack-other.c
OphirMunk: type fixed

lib/conntrack.c
OphirMunk: I prefer leaving get_tm outside of get_hw_active mainly because get_hw_active should only check HW activity and not doing extra. We can however do something like: 
```
if (get_tm() && conn_hw_active())
```
Additionally please note that get_tm always return 0 in current implementation so you should not concern that tm is not valid.

lib/conntrack.c
OphirMunk: Moved

lib/conntrack.c
OphirMunk: Indeed if there are no calls for dump-conntrack in the background we are not called here. Bug on the other hand we are not called either soon enough under sweep. This requires more investigation.

lib/conntrack.c
elibritstein: the concern is just for code correctness. in the future someone might implement a new protocol that returns -1 for get_tm...

lib/conntrack.c
elibritstein: it's #define CT_TM_MIN (30 * 1000)                                                  
but i think it should be ok in the place i put it. please check it

PR #89, opened 2019-10-31
Output upstream by eli

OphirMunk:
Pushed from gerrit as a temporary work around for everyone to review.


OphirMunk:
A new PR with latest updates is planned

